PR #24, opened 2016-10-18
Corrected spelling of "Glycol" and defined the VOX switch for CDR and…

indy91:
… LMP audio.


indy91:
You forgot to create the actual pull request, so I did that for you. You renamed the Glycol rotational switch, but I think you forgot to commit your changes to LEM.h, which still will have the line "RotationalSwitch ClycolRotary;". You can add commits to pull requests, should be no problem.

indy91:
Needs some more fixes it seems :D

rcflyinghokie:
Looks like everything is happy now :)

indy91:
I did a check in the saturnpanel file, because I wasn't sure if the switch name has to be unique in LM and CSM, which could have been a reason for it to be named Clycol instead of Gylcol. But both CSM and LM have abort switches with the name "AbortSwitch", so I don't think that is the case. So I think it is safe to be merged.

PR #25, opened 2016-10-24
Initial LM CB Configuration

rcflyinghokie:
This changes the initial switch/cb configuration for the LM to match that of the initial switch status from the Apollo 12 activation checklist.


indy91:
Looks good.

PR #26, opened 2016-10-24
Fix CB Spelling Error: "AQS" CB to "AGS" CB

rcflyinghokie:
Panel 11 cb "AQS" should actually be "AGS"


indy91:
Yep, even the bitmap has this wrong wrong.

PR #27, opened 2016-10-25
Changed switch spelling "PNGS" to "PGNS"

rcflyinghokie:
This change makes all of the LM switches/breakers use "PGNS" as the spelling.  They were mis-matched with some using "PNGS" and some using "PGNS"


indy91:
What about adding the thing you actually want to be commited? :D Maybe it got eaten in the merge or so.

rcflyinghokie:
Did it not work?

indy91:
There were no change files before. Now it looks good.

rcflyinghokie:
Doh, I forgot to commit before pushing :P

indy91:
Also don't forget to merge the merge. After I have merged this pull request, your master branch will be one commit behind. There is no actual difference in the files, just the merge commit is what is missing. So you have to fetch upstream and then merge it into your master branch. I don't think it is a problem, just the "clean" way to synchronize the two master branches.

PR #29, opened 2016-11-23
Apollo 7 & 8 Checklist dv switch positions and minor grammatical fixes

rcflyinghokie:
Apollo 8 now launches with the CG switch on LM/CSM as per the AOH, gives the option to switch to CSM in the P40 Checklist.  Apollo 7 minor fix re-guards tower jett switches automatically during boost.  Fix Flightplan group start time so it starts after CSM/LV SEP instead of right after boost.  Minor grammatical/verbiage fixes.


indy91:
Thanks.

PR #31, opened 2016-11-27
Added Fail option to P51/52 Alignment to reject a star angle.

rcflyinghokie:
Fixed some copy/paste errors in the P51/52 checklists as well as added a fail option to reject a bad star angle.


rcflyinghokie:
Disregard this pull.

PR #44, opened 2017-01-30
Slight standby implementation changes

eddievhfan1984:
-Adjusted standby debug line position (otherwise requires optics on)
-Added reset of SbyPressed state

Signed-off-by: Kyle Tekaucic <tekaucic.2@buckeyemail.osu.edu>

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_sys/yaAGC/agc_engine.c
+++ b/Orbitersdk/samples/ProjectApollo/src_sys/yaAGC/agc_engine.c
@@ -1712,6 +1712,12 @@ agc_engine (agc_t * State)
       return (0);
     }
   
+  // Same with Standby
+  if (0 == (State->InputChannel[032] & 020000))
+	  State->SbyPressed = 1;
+  else
+	  State->SbyPressed = 0;
+
   //----------------------------------------------------------------------
   // Here we take care of counter-timers.  There is a basic 1/3200 second
   // clock that is used to drive the timers.  1/3200 second happens to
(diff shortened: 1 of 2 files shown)


Orbitersdk/samples/ProjectApollo/src_sys/yaAGC/agc_engine.c
```diff
@@ -1749,6 +1749,8 @@ agc_engine (agc_t * State)
 		  // Same with Standby
 		  if (0 == (State->InputChannel[032] & 020000))
 			  State->SbyPressed = 1;
+		  else
+			  State->SbyPressed = 0;
```
thewonderidiot: I recommend moving this check outside of the `while (ScalerCounter >= SCALER_OVERFLOW)` loop. Placing it here makes the check for PRO release only occasionally happen based on the current scaler time; in reality there's no clocks gating the reset so it happens immediately upon release. So you might get some slightly weird timing effects. Putting the check to do this just before the while loop should be fine.

eddievhfan1984: Done in the new commit.

indy91:
Does the change look good to you @thewonderidiot ? I wouldn't know...

thewonderidiot:
Ah, sorry, what I meant is that the setting and unsetting can't be in the same spot. The original code was in the right place, as PRO being depressed is only checked for by the hardware once every 1.28 seconds. It's just the release that is continuously monitored.

The code should be something like this:

```
if (1 == (State->InputChannel[032] & 020000))
   State->SbyPressed = 0;

...

while (ScalerCounter >= SCALER_OVERFLOW)
     {

...

        if (0 == (State->InputChannel[032] & 020000))
 		State->SbyPressed = 1;

...
```

indy91:
Looks good to me now.

PR #45, opened 2017-02-01
Minor PRO release fix

thewonderidiot:
I was apparently too sleepy when I gave the code example -- the condition as written is impossible. Whoops! This should work correctly now.


indy91:
Thanks!

PR #55, opened 2017-03-07
Add Apollo 9 Checklist Locations

rcflyinghokie:


indy91:
Looks good, except for the "Apollo 9 Checklists.xls" file. I don't think we actually have that file yet, so if you created one, then you should add it to this pull request.

indy91:
Uhh, I have an Apollo 9 checklist in my local copy. Did you create that? In any case, just add it to the pull request I will commit it.

rcflyinghokie:
I created both checklist files when I did my initial test with git and the 2015 branch, that and the LM Default Checklists.

indy91:
That was 2 days ago! You can't expect me to remember that!

Wow...

PR #56, opened 2017-03-11
CM independence

eddievhfan1984:
This is the beginnings of an independent CM. At the moment, I have the project unloaded in the VS solution so as to not auto-compile and give the build-checker a hassle.


indy91:
The auto check doesn't seem to like it. Let's see what the problem is...

eddievhfan1984:
According to my VS build, the issue is targeted in the vAGC files:

`Severity	Code	Description	Project	File	Line	Suppression State
Error	C1189	#error:  Sorry, cannot determine the target operating system.	CM	d:\orbiter 2016\orbitersdk\samples\projectapollo\src_sys\yaagc\yaAGC.h	73	
Error	C4996	'fileno': The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _fileno. See online help for details.	CM	D:\Orbiter 2016\Orbitersdk\samples\ProjectApollo\src_sys\yaAGC\agc_engine_init.c	212	
`

indy91:
Are the project properties configured like in the other projectes? All the NASSP specific changes for that?

indy91:
It might not like the C files among all the C++ stuff. The Saturn project files might be configured for that in some way.

eddievhfan1984:
In that case, it might be better to just close the request while I figure out what the hell is going on.

What really stinks is that apparently there aren't "solution default" properties I can import when create a new project, or if there are, I can't find them. >.<

PR #64, opened 2017-04-01
Make power transients work in standby

Thymo-:
I noticed the AGC didn't properly restart when a power transient occurs when the agc is in standby.
Because the program counter is set to 4000 by agc_engine upon entering standby, the hardware is never restarted.
This change removes that check, resets the standby flip-flop and turns the DSKY PSU back on.

Thanks to @thewonderidiot for clarifying along the way.


Thymo-:
Reference: #73

PR #67, opened 2017-04-05
STBY and RESTART are both non latching and should turn off on transient

Thymo-:
This pull request is an addition to #64 where I made some standby mode power transient changes.
The STBY and RESTART light are actually non latching and should be turned off on a power loss.


Thymo-:
I made a typo in a comment. I'll fix it tomorrow, don't merge this yet please.

Thymo-:
All is good now.

Thymo-:
Reference: #73

Issue #13, opened 2015-09-02
Systems Modelling Improvement: Phase 1


dseagrav:
Systems modelling to enable failures and breakage due to mismanagement.

PR #71, opened 2017-04-12
Implement Timer corruption and power draw

Thymo-:
This pull request adds PanelSDK support to the timers in the CSM and LM. Also implemented is a new feature that scrambles the time flip flops to a random value when power is lost.

The Mission and Event Timers in the CSM are wired to the TIMERS MNA and MNB breakers on panel 229. These can operate independently from the display which runs of off the two Numerics dials (timer on MDC to LMDC numerics, timer on 306 to LEB numerics) on AC.

The power draw values where pulled from the [CSM-104 Systems Handbook](https://www.ibiblio.org/apollo/ApolloProjectOnline/Documents/CSM%20104%20Systems%20Handbook%2019690124.pdf) for the timers. The display power usage code has been borrowed from the DSKY (assuming all segments lit). @indy91 has provided me with values for the LEM timers. The event timer also consumes more power when it is running because it has mechanical counters.

These changes are fully backwards compatible with older scenarios. All required values will be saved on the next simulation exit.

This PR helps towards achieving #13 and #6.


indy91:
Thanks!

Thymo-:
Reference: #73

PR #72, opened 2017-04-13
CWS Fixes and Tweaks

Thymo-:
@jalexb88 mentioned there were some differences between our simulation behaviour and in this [video](https://youtu.be/oxgGm8kHuig). From this video it can be clearly seen that as soon as the V35 is run the Master Alarm comes on. It also does not blink.

After discussing this in the IRC channel and examining the related schematics it was concluded that the master alarm is not supposed to blink nor that a light test causes the master alarm to be inhibited.

Our code also checks for the TestAlarms bit (ch13 bit10) in the AGC to be set to determine if a light test is in progress. @thewonderidiot pointed out that this signal is in fact completely internal to the AGC and should not be able to be read by the CWS. This functionality is being retained for now because agc_engine doesn't control the CMC light yet. A comment has been added that it should be removed when this gets added.

Some tweaks have been made to the power consumption values as the previous ones were a bit exaggerated. The consumption has been changed from 2W to 0.6W. This is the same value that is used for the DSKY lamps and might be changed again if better values are found.

Also included is a slight change to the mission timer code. I forgot to adjust the power draw value for the LEM timer.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/csmcautionwarning.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/csmcautionwarning.cpp
@@ -247,15 +247,10 @@ void CSMCautionWarningSystem::TimeStep(double simt)
 	// ISS warning
 	//
 
-	if (aws.ISSWarning) {
+	if (aws.ISSWarning)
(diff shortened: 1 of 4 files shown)


Thymo-:
Reference: #73

PR #91, opened 2017-05-23
Apollo 9 Checklists Through TPI

rcflyinghokie:
TPI procedures still needs some modification and cleaning up, this PR is mostly to make sure my git is still properly set up after the change to Orbiter2016 branch.


indy91:
Seems like it's properly set up!

PR #93, opened 2017-05-27
Add RCS CMD delay and remove abort light switch from all checklists.

rcflyinghokie:


indy91:
Great that you got it working again!

PR #105, opened 2017-06-13
Saturn Mass Changes, Apollo 10 IECO, Add Saturn Mass Table

rcflyinghokie:


indy91:
You also commited your changes to LVDC.cpp, if I merge this then every Saturn V flight would have an early IECO of the S-II. I will do the required LVDC changes myself, so that the S-IC and S-II IECO times and if an S-II IECO even happens can be chosen as scenario parameters.

rcflyinghokie:
Ah ok no problem.  I thought it had an Apollo 10 and up line in there.  Also the IECO doesn't work on time anyways, but it does work.  I think there is also a fuel condition not just time.

PR #147, opened 2017-08-17
Add ED VOLTS Off initialization

Thymo-:


Thymo-:
Doesn't work as expected. Further testing needed.

PR #155, opened 2017-09-18
Apollo 9 Checklists through SPS4, remove "flightplan" group from default checklists.

rcflyinghokie:


rcflyinghokie:
Also added a V16N72 for an Apollo 11 rendezvous radar adjustment

PR #168, opened 2017-10-10
AGS Testing

thewonderidiot:
(Don't merge me!)

This is the first set of tests to try out possible fixes for AEA misbehaviors, as discussed in virtualagc/virtualagc#1052. This first attempt contains very similar instruction tweaks:
1. `COM 400000` now sets the overflow indicator.
2. `ABS 400000` now sets the overflow indicator.

The self-tests still pass with both of these tweaks in place, so at the very least it doesn't seem to break anything fundamental. If this doesn't work or has bad effects on mission software, we should try it with only the ABS change, and then with only the COM change.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_lm/yaAGS/aea_engine.c
+++ b/Orbitersdk/samples/ProjectApollo/src_lm/yaAGS/aea_engine.c
@@ -128,6 +128,10 @@ static const int64_t CONST64_3 = 1i64;
 // rounded quotient is captured as one of the output parameters.  This 
 // calculation assumes that "long long" is at least 64 bits.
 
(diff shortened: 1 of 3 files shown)


thewonderidiot:
I've added in the bugfix for the LLS instruction that I described in the virtualagc issue.

thewonderidiot:
@rcflyinghokie, @indy91, I've put in the ABS and COM testing code that I was grumbling about earlier today. Would you mind running through some typical AGS scenarios while watching the abscom_problem_addr variable I added? If it ever becomes non-zero, jot down the address. That would mean that the ABS/COM changes did have some effect on program flow, and give us a place to look at to determine whether that effect was good or bad.

PR #285, opened 2018-03-04
IMUCase was uninitialized, added to constructor.

Thymo-:
Since I was building in debug mode variables get initialized to a non-null value by the memory allocator. This uncovered an uninitialized variable in the IMU causing a CTD. I've added it to the constructor to explicitly null it,

I recommend everyone that works on the code to build their modules in debug mode to find issues like these.

@indy91


indy91:
Good find, thanks!

PR #426, opened 2018-12-15
Ignore XRSound Addon

msligo:
I use the XRSound addon with Orbiter2016. When I created a Project Apollo git repo in my Orbiter folder, I noticed that it doesn't ignore the XRSound files, so I added them to the .gitignore file.

I'm new to git/github, so if I've done anything wrong please let me know.


indy91:
Looks right. Thanks!

PR #449, opened 2019-01-09
Fixing h_Tank area calculation

msligo:
These two lines are clearly trying to calculate the cross sectional area of a spherical tank from its volume, however the equations were incorrect. This commit corrects them.

This commit also slightly improves the efficiency of H_system::Build() by allowing it to skip NULL lines in the config file.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_sys/PanelSDK/Internals/Hsysparse.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_sys/PanelSDK/Internals/Hsysparse.cpp
@@ -420,8 +420,9 @@ void H_system::Build() {
 			Create_h_WaterSeparator(line);
 		else if (Compare(line, "<HEATLOAD>"))
 			Create_h_HeatLoad(line);
-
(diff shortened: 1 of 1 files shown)


indy91:
Both changes look good. But changing the tank area like that (the difference seems quite significant, the area is now about 10x smaller for an average tank size) might have lots of implications for the thermal behavior of tanks. I'll first check if that could cause any problems. The isolation factors of many tanks might have been fine tuned to the old and wrong area. Then I'll merge this PR.

But thanks for working on this! There is definitely a lot than can be done to improve the Panel & Systems SDK, especially for the LM ECS.

indy91:
As this will make the thermal area for tanks smaller rather than bigger, I think it won't lead to any issues. Maybe some tanks that are expected to heat up or cool down due to Sun exposure (or lack thereof) need some tweaks in the future. We will see.

msligo:
Of course, I'm sorry, I should have thought of that. I don't want to create more work for you, so in future I'll be sure to fully investigate the impact of each change before submitting a pull request.

PR #456, opened 2019-03-02
Fix and clean up .gitignore

Thymo-:
The way the gitignore was setup all source files that were not already
tracked would be ignored. If a subdirectory of a directory is ignored it
can not be negated unless all parent directories are excluded too. The
other dirs should be ignore with $dir/*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,7 +34,6 @@ _$*
 *.elc
 *.ln
 *.hgtags
-core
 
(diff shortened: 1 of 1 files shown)


Thymo-:
I also removed the ignoring of "core". Normally this is the memory dump done by yaAGC every 15 seconds but we disabled that at some point.

Thymo-:
Ignored core files again. They are still generated if done through PAMFD.

PR #461, opened 2019-04-29
Fix out of range access violation

schnepe2:
- increased LVDC1B::SSTTB size, as access in LVDC1B::TimeStep (LCDC.cpp line 4785) is out_of_range.
- added comments in header to clarify usage.


indy91:
Oops, that's a pretty embarrassing bug. Thanks for findind it!

schnepe2:
No worries. Thanks for merging ;)

PR #468, opened 2019-05-15
New mouse click handling in class RotationalSwitch for more realistic…

jalexb88:
… behavior. No longer can you go directly to any selection around the rotary, you must sequence through every position to get to desired selection.

- added a flag to specify is rotary can wraparound (infinite rotation)
- modified panel rotaries (and associated logic if applicable) where needed for compatibility with the new function.
- added wraparound flag where needed


jalexb88:
Issue # 315

PR #512, opened 2019-09-29
Replace OrbiterSound with XRSound

### LINKED ISSUE #474: Implement XRSound
As we know, OrbiterSound 4.0 is not officially compatible with Orbiter 2016. While it does run without too many issues, there are some good reasons to make the switch, such as (not limited to) these issues: 

- 3D audio causing low frame-rate at session start until a sound is played. 
- Weird rumbling noise heard around the time of pitch-over during P64 . 
- Radio/mp3 Panel MFD causes a CTD

XRSound: https://www.alteaaerospace.com/ is the best candidate right now for a new sound solution for NASSP. Currently OrbiterSound is deeply intertwined in NASSP code so it will require a fair amount of work to successfully transition over to XRSound but the in the long run this should be good sound platform for NASSP 8.0+

ethhics:
**This PR is NOT complete! Don't merge yet.**

Closes #474 .

I want to provide a starting point for y'all before this just sits on my hard drive. Feel free to commit to this branch as needed to finish the implementation.

**What is implemented**

- [X] `src_sys/soundlib` has a complete implementation of XRSound in a way that should require minimal modification of any other code.
- [X] Build files have been changed to target XRSound.

**Still to be done**

- [ ] Vessels load and initiate connection to `XRSound.dll` before `XRSound.dll` is loaded. This results in nothing being configured from a sound perspective. It appears that the vessels' `soundlib` code will need to be shifted around so that *nothing* happens before `clbkPostCreation`.
- [ ] Someone more intimately familiar with the project should do a runthrough to make sure that all of the sounds that are supposed to play, play.
- [ ] It would be "The Right Thing" to tear down `soundlib` and everything related to sounds playing and rewrite it in the object-oriented style of XRSound and take advantage of features like >60 samples. I'm unfamiliar with this project's code philosophy, so if that's more of the approach y'all'd like, just close this PR and I'll devise a strategy for a more thorough codebase surgery.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/Build/VC2017/ApolloRTCCMFD.vcxproj
+++ b/Orbitersdk/samples/ProjectApollo/Build/VC2017/ApolloRTCCMFD.vcxproj
@@ -85,7 +85,7 @@
     </Midl>
     <ClCompile>
       <Optimization>Disabled</Optimization>
-      <AdditionalIncludeDirectories>../../../LaunchMFD/libMFD;../../src_aux;../../src_sys;../../src_mfd;../../src_lm;../../src_moon;../../src_csm;../../src_launch;../../src_landing;../../src_saturn;../../../../include;../../../../../Sound/OrbiterSound_SDK/VESSELSOUND_SDK/ShuttlePB_project;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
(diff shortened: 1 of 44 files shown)


indy91:
Thanks for working on this. Looks like a big project. I can't contribute much at the moment, I'm too deep into working on some RTCC MFD stuff. How about you continue on the path you are right now with this PR and try to get some things working, I think it definitely is on the way to be able to be merged at some point. If you want to talk about some of the details or have any questions concerning the NASSP code, feel free to stop by in our IRC channel (#nassp on freenode), that's where we usually hang out.

ThymoNL:
This is still on the roadmap and it would be nice to have for NASSP 8, this does however need rework to be compatible with the current state of NASSP and it's probably easier to redo these changes from the latest state.

PR #521, opened 2020-04-26
Csm antennas

n7275:
I'm submitting a pull-request for the CSM HGA functionality.
 
I'm 98% sure I got everything this time.

All modes appear to be working. was able to set manual position in lunar orbit with the controls, and then reacquire automatically at earth-rise

I have tested REACQ mode through 5 lunar orbits and it works flawlessly on my end.

Scan-limit warning and scan limit are working now. the antenna switches back to wide mode after hitting the warning and switches back to manual(in reaqc mode). in auto mode the antenna "stops tracking" on scanlimit which is correct per CSM Systems handbook.

The switch, tracking mode, and scanlimit logic, and the actual tracking control logic are now separated. tracking control could probably be a function, but it would need 12 arguments so I've left it for now.


indy91:
Or rather some duplications, like the Fra Mauro config. Once named "Fra Mauro" and once "FraMauro". Not sure how that would have happened.

PR #529, opened 2020-05-17
LC39 mesh, ML mesh and ML animations update.

Jordan-64:


indy91:
Looks good so far. What is "MetalPlatte - Kopie.dds" and "MetalPlatte_norm - Kopie.dds". Are those needed or accidental duplications?

Jordan-64:
These are just variations that are actually no longer needed and can be deleted.

PR #540, opened 2020-08-22
Rrt connectors

n7275:
I think this is at a point where it can be added. There are a few minor internal system things like the tone generator/filter heater. Because I don't have a range-accuracy filter-temperature-dependence model, and we're nor simulating errors in other RF systems, at the moment, I think this should wait.


n7275:
@indy91 I made a few last minute updates, the most notibly of which was to remove skin-tracking functionality. It's easy to add back in, but I can't find positive confirmation that it could be done with the LEM RR.

indy91 (approved):

PR #549, opened 2020-11-01
Scaler emulation in AGC is changed to accomodate 12800pps heart beat for IO devices

ggalfi:
Should not change the behaviour of AGC. Two new scenario param is defined (SCALER, SCALERCHANGED). Compatible with earlier scenarios.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_sys/apolloguidance.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_sys/apolloguidance.cpp
@@ -429,6 +429,8 @@ void ApolloGuidance::SaveState(FILEHANDLE scn)
 	oapiWriteScenario_int (scn, "IDXV", vagc.IndexValue);
 	oapiWriteScenario_int (scn, "NEXTZ", vagc.NextZ);
 	oapiWriteScenario_int (scn, "SCALERCOUNTER", vagc.ScalerCounter);
+	oapiWriteScenario_int(scn, "SCALER", vagc.Scaler);
+	oapiWriteScenario_int(scn, "SCALERCHANGED", vagc.ScalerChanged ? 1 : 0);
 	oapiWriteScenario_int (scn, "CRCOUNT", vagc.ChannelRoutineCount);
 	oapiWriteScenario_int(scn, "DSKYCHANNEL163", vagc.DskyChannel163);
 	oapiWriteScenario_int(scn, "WARNINGFILTER", vagc.WarningFilter);
@@ -462,6 +464,10 @@ void ApolloGuidance::LoadState(FILEHANDLE scn)
 	//
 	// Now load the data.
 	//
+
+	boolean  //Flag to keep compatibility with older scenario files.
+		scalerLoaded = false 
+		;
 
 	while (oapiReadScenario_nextline (scn, line)) {
 		if (!strnicmp(line, AGC_END_STRING, sizeof(AGC_END_STRING)))
@@ -505,6 +511,15 @@ void ApolloGuidance::LoadState(FILEHANDLE scn)
 		}
 		else if (!strnicmp (line, "SCALERCOUNTER", 13)) {
 			sscanf (line+13, "%d", &vagc.ScalerCounter);
+		}
+		else if (!strnicmp(line, "SCALER", 6)) {
+			sscanf(line + 6, "%d", &vagc.Scaler);
+			scalerLoaded = true;
+		}
+		else if (!strnicmp(line, "SCALERCHANGED", 6)) {
+			int tmp;
+			sscanf(line + 6, "%d", &tmp);
+			vagc.ScalerChanged = tmp;
 		}
 		else if (!strnicmp (line, "CRCOUNT", 7)) {
 			sscanf (line+7, "%d", &vagc.ChannelRoutineCount);
@@ -567,6 +582,9 @@ void ApolloGuidance::LoadState(FILEHANDLE scn)
 		papiReadScenario_bool(line, "PROGALARM", ProgAlarm);
 		papiReadScenario_bool(line, "GIMBALLOCKALARM", GimbalLockAlarm);
 	}
+
+	if (!scalerLoaded)
+		vagc.Scaler = (vagc.InputChannel[ChanSCALER1] << 2) | (vagc.InputChannel[ChanSCALER2] << 16);
 }
 
 //
--- a/Orbitersdk/samples/ProjectApollo/src_sys/yaAGC/agc_engine.c
+++ b/Orbitersdk/samples/ProjectApollo/src_sys/yaAGC/agc_engine.c
@@ -1847,7 +1847,7 @@ SimulateDV(agc_t *State, uint16_t divisor)
 // it requires slightly more processing, in order to conform as obviously
 // as possible to the MIT docs.
 
-#define SCALER_OVERFLOW 80
+#define SCALER_OVERFLOW 20
 #define SCALER_DIVIDER 3
 
 // Fine-alignment.
@@ -1870,7 +1870,7 @@ static unsigned ImuChannel14 = 0;
 int
 agc_engine (agc_t * State)
 {
-  int i, j;
+  int j;
   uint16_t ProgramCounter, Instruction, /*OpCode,*/ QuarterCode, sExtraCode;
   int16_t *WhereWord;
   uint16_t Address12, Address10, Address9;
@@ -1905,6 +1905,15 @@ agc_engine (agc_t * State)
   //  }
 
   State->CycleCounter++;
+
+  //----------------------------------------------------------------------
+  // Update the thingy that determines when 1/12800 second has passed.
+  // 1/12800 is the basic timing used to drive timer registers.  1/12800
+  // second happens to be 20/3 machine cycles.
+
+  State->ScalerCounter += SCALER_DIVIDER;
+  State->ScalerChanged = 0;
+  State->DskyTimer += SCALER_DIVIDER;
   
   //----------------------------------------------------------------------
   // The following little thing is useful only for debugging yaDEDA with
@@ -1925,16 +1934,6 @@ agc_engine (agc_t * State)
       ShiftToDeda (State, (Data >> 3) & 7);
       ShiftToDeda (State, Data & 7);
     }
-
-  //----------------------------------------------------------------------
-  // Update the thingy that determines when 1/1600 second has passed.
-  // 1/1600 is the basic timing used to drive timer registers.  1/1600
-  // second happens to be 160/3 machine cycles.
-
-  State->ScalerCounter += SCALER_DIVIDER;
-  State->DskyTimer += SCALER_DIVIDER;
-
-  //-------------------------------------------------------------------------
 
   // Handle server stuff for socket connections used for i/o channel
   // communications.  Stuff like listening for clients we only do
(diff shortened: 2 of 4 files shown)


Orbitersdk/samples/ProjectApollo/src_sys/yaAGC/agc_engine.c
```diff
@@ -1905,6 +1905,15 @@ agc_engine (agc_t * State)
   //  }
 
   State->CycleCounter++;
+
```
Thymo-, line 1908: Are these changes also committed to virtualAGC? We should not start diverging from upstream again.

Orbitersdk/samples/ProjectApollo/src_sys/apolloguidance.cpp
```diff
@@ -463,6 +465,10 @@ void ApolloGuidance::LoadState(FILEHANDLE scn)
 	// Now load the data.
 	//
 
+	boolean  //Flag to keep compatibility with older scenario files.
```
Thymo-, line 468: Put this on a single line like `boolean scalerLoaded = false; // Flag to keep compatibility with older scenario files.`

Orbitersdk/samples/ProjectApollo/src_sys/apolloguidance.cpp
```diff
@@ -567,6 +582,9 @@ void ApolloGuidance::LoadState(FILEHANDLE scn)
 		papiReadScenario_bool(line, "PROGALARM", ProgAlarm);
 		papiReadScenario_bool(line, "GIMBALLOCKALARM", GimbalLockAlarm);
 	}
+
+	if (!scalerLoaded)
```
Thymo-, line 586: Wouldn't this also work by checking if vagc.Scaler == 0? This would avoid introducing a new variable.

Orbitersdk/samples/ProjectApollo/src_sys/yaAGC/agc_engine.c
```diff
@@ -1847,7 +1847,7 @@ SimulateDV(agc_t *State, uint16_t divisor)
 // it requires slightly more processing, in order to conform as obviously
 // as possible to the MIT docs.
 
-#define SCALER_OVERFLOW 80
+#define SCALER_OVERFLOW 20
```
Thymo-, line 1850: We should not be diverging from VirtualAGC, it will only cause problems and frustration in the future. If a change to yaAGC is needed a PR should be made in https://github.com/virtualagc/virtualagc.

Orbitersdk/samples/ProjectApollo/src_sys/yaAGC/agc_engine.c
```diff
@@ -1870,7 +1870,7 @@ static unsigned ImuChannel14 = 0;
 int
 agc_engine (agc_t * State)
 {
-  int i, j;
+  int j;
```
Thymo-, line 1873: Can be renamed to i as j is used for nested loops.

Orbitersdk/samples/ProjectApollo/src_sys/yaAGC/agc_engine.c
```diff
@@ -2012,266 +2010,270 @@ agc_engine (agc_t * State)
...
+			  }
+			  else if ((State->RuptLock || State->NoRupt) && 0300 == (0777 & chscal1))
+			  {
+				  // We've either had no interrupts, or stuck in one, for 140ms. Sound the alarm!
+				  TriggeredAlarm = 1;
 
-			  // Set the RUPT LOCK bit in channel 77.
-			  State->InputChannel[077] |= CH77_RUPT_LOCK;
-		  }
+				  // Set the RUPT LOCK bit in channel 77.
```
Thymo-, line 2122: check whitespace

ThymoNL (changes requested):
I missed the discussion leading to these changes so I'm not able to fully understand why these changes are needed. Could you please elaborate a bit on what you're trying to achieve? :)

There are quite a lot of changes to yaAGC. They should not be changed in NASSP but in VirtualAGC as that is our upstream. When the changes have been included there we can update our files to sync with their repo. Doing so also means these changes can be reviewed by more people.

Tip: It's best practice to always split your code into small commits detailing why that change was made. This not only makes it easier for others to look at your code but also for yourself to keep track of what you changed and to roll back to a previous state in case something breaks. It's a lot of work to go back and change this, so don't bother. Please do keep it in mind for next time though. :-)

Orbitersdk/samples/ProjectApollo/src_sys/yaAGC/agc_engine.c
ggalfi: > 
> 
> I missed the discussion leading to these changes so I'm not able to fully understand why these changes are needed. Could you please elaborate a bit on what you're trying to achieve? :)
> 
> There are quite a lot of changes to yaAGC. They should not be changed in NASSP but in VirtualAGC as that is our upstream. When the changes have been included there we can update our files to sync with their repo. Doing so also means these changes can be reviewed by more people.
> 
> Tip: It's best practice to always split your code into small commits detailing why that change was made. This not only makes it easier for others to look at your code but also for yourself to keep track of what you changed and to roll back to a previous state in case something breaks. It's a lot of work to go back and change this, so don't bother. Please do keep it in mind for next time though. :-)

The desired final state would be similar what you can see in the agcio branch (not this agcio2 ) in my repo. On agcio I have written a post in May where I summarized the goals of this development, you may read it here: https://www.orbiter-forum.com/threads/simulation-of-apollo-11-lunar-landing.37948/ . The main reason why I have to touch the AGC engine itself is that it should interact with most of the peripherials (CDUs, PIPAs, radars, etc.) in a synchronized way for proper simulation. The original goal was to create an accurate simulation of Apollo 11, but as a side effect it turned out that the accuracy of the PIPAs could be improved significantly by this. I also have some ideas on how the accuracy and fuel-efficiency of the DAP could be improved (or more precisely: to entirely utilize the "steady as rock" feature already provided by Colossus and Luminary softwares) but to be able to do so I'll also need this synchronous behaviour of AGC. We had a discussion on IRC with Mike and Indy (see the logs about from 2 weeks ago) on how to integrate all these stuff into the NASSP main branch. What came out of this was that the top priority is not to ruin any functionality which is available now in NASSP. My agcio branch currently fails this in two aspects: 1. it cannot raise hw alarms 2. not works well in multithreaded mode, but none of this seem to me a unsolvable issue, I just simply haven't taken care of them in the pervious development cycle. However making these changes compatible with VirtualAGC is just a nice-to-have thing at the moment. Anyway I've considered to make changes on VirtualAGC first and then migrate them into NASSP. The problem is that agc_engine.c in VirtualAGC has been _already_ diverged from the one in NASSP version, and I cannot see any reliable method to do this "NASSPsifying" after I make a change in VirtualAGC.

To make things clear, my roadmap for this transition is the following:
1. Increase the resolution of the scaler by two bits (or implement at all the scaler if you will), because that is the depth  required by peripherials. It could have been implemented outside of agc_engine.c but firstly I think as the scaler circuit is one of the core part of the real AGC hw so it's simulation should go into the emulator itself. Secondly, the agc_engine is doing some scaler mgmt already, if we repeat it outside of agc_engine, it will have impact on the performance. These steps are running 100k times in every simulated second, so in 100x acceleration they could have sensible impact.
2. Adding the properly implemented counter-register functionality to agc_engine. I intend to do this in a way that the AGC still remains compatible with the current (asynchronous) NASSP peripherials, but will be able accomodate synchronous devices as well. Will require some care here in case of outgoing registers, as they are behaving differently in the current emulation than on the real hw. In the worst case there will be some "#ifdef REAL_CDUERR_IMPL" parts in the code.
3. Hopefully after step 1-2 there will not be further need to modify agc_engine, and the synchronous and maybe more physics-oriented implementation of peripherials will happen in this step. I have done already this more-or-less completely for CDUs, the IMU gimbals, IMU coarse alignment, the PIPAs (I'm fairly proud of the in-depth but still surprisingly fast physical pendulum simulation there :) ), RR and optics angles and coarse alignment, LR and RR data readout, FDAI errors, CSM TVC/SIVB stuff, LM thrust drive. I haven't touched but they are working in their original (asynchronous) way: IMU Gyro Torquing, Telemetry Up and Downlink. To see what I intend to wire in, please refer to the commits of the agcio branch. In the best case  only the multithreading issue will inflict some changes till my codes make their way into agcio2.

So this PR corresponds to step 1.

Orbitersdk/samples/ProjectApollo/src_sys/apolloguidance.cpp
ggalfi: Yeah there will more "something is loaded" variables will arrive here. That's why the multiline declaration.

Orbitersdk/samples/ProjectApollo/src_sys/apolloguidance.cpp
ggalfi: One thing what I learned during my career that if I don't want to cause terrible headache for myself or for any other poor bastard then I shouldn't use 0 as a missing value :) Additionally as I mentioned in my previous response, there will be other new scen paramteres like CDU counter registers which I plane to read in here as well, and there getting a 0 as loaded value is much more likely than in case of the scaler, so with this trick we could mess up the saved simulation state. And apart from that: why does the introduction of a new local variable like this means any problem? It takes max. 4 bytes from the stack in a code which runs only once in a session so no measurable impact on performance, and it also makes the intention of the programmer much more clearer.

Orbitersdk/samples/ProjectApollo/src_sys/yaAGC/agc_engine.c
ggalfi: See my general response above. I'm open to start with modifying first VirtualAGC - actually even I find it the neatest way - but then please help me out with a method on how to transform the VirtualAGC-compatible agc_engine to NASSP-compatible one for testing. Maybe creating a patch between the two "official" versions in NASSP and VAGC, and applying it to my developments will do the job... But I'm really curious on what is the best practice here.

Orbitersdk/samples/ProjectApollo/src_sys/apolloguidance.cpp
Thymo-: I got the upstream yaAGC to run on NASSP today. Still need to test that everything works as it should but looks to be working so far.

ggalfi: > I got the upstream yaAGC to run on NASSP today. Still need to test that everything works as it should but looks to be working so far.

That's really great, thanks! Am I supposed to use the sync_agc (seem to me the same as fix_agc)? During the weekend I'll build up something similar to this PR but on top of sync_agc and submit it. Is Do you concur?

Thymo-: There is still some work to be done. I didn't manage to do a landing with the new AGC yet (partially because I don't know how to fly them LEM yet) so there might still be some issues.
Then yaAGS also needs another look as it currently depends on yaAGC to build. It would be easier to separate those before introducing new changes.

What we can do is once this stuff is mainlined: adapt your PR to work with it and then test it in a feature branch. If everything is working as it should you can submit those changes to VirtualAGC to get them upstreamed. That way we won't diverge again. I'll make another commit today with some final fixes so you can get started on porting it over. The rest of the changes to come will be minor so shouldn't present problems with getting your work integrated.

Issue #551, opened 2020-11-02
Work with virtualagc devs to move to upstream yaAGC


Thymo-:
Our counterpart of https://github.com/virtualagc/virtualagc/issues/622
Work is being blocked by us not being in sync with upstream yet.
Use this as a tracker bug for any effort towards unification.

The sync_agc branch contains the upstream files, it does not currently build.
The fix_agc branch contains fixes to make sync_agc build and to test individual patches. This branch may be synced to sync_agc and Orbiter2016 when desired.

PR #554, opened 2020-11-04
Project changes and some AGC cleanup

Thymo-:
Cleaned up some dead code in apolloguidance. The functions in agc_engine are still there and not used. Those will be removed when we move to upstream (#551).

Ignored and removed all vcxproj.user files as these are user-specific overrides to project settings and should not be in source control.
Also added _CRT_SECURE_NO_WARNINGS (new version of _CRT_SECURE_NO_DEPRECATE) to Preprocessor Options so we don't have to clutter our code with it (and edit virtual agc files).


Orbitersdk/samples/ProjectApollo/src_sys/apolloguidance.cpp
```diff
@@ -836,41 +836,31 @@ void ApolloGuidance::ProcessIMUCDUReadCount(int channel, int val) {
 }
 
 void ApolloGuidance::GenerateHandrupt() {
-	GenerateHANDRUPT(&vagc);
+	vagc.InterruptRequests[10] = 1;
```
thewonderidiot, line 839: So basically you're just copying the bodies of these functions out of agc_engine.c, so you can later delete the functions from there? Seems reasonable.

PR #561, opened 2020-11-22
Make fuel cells' output voltage drop with increasing load.

n7275:
Internal resistance is taken as 0.0346667ohms which gives approximately predicted results(27 volts at 1420 watts), and you can get bus under-volt alarms by drawing more than the max on a single cell, as long as you don't draw enough to trip the FC disconnect. Any systems that are expecting a min voltage of exactly 28.8 V may be effected, but if they are set to the specs from the systems handbook they should behave as expected.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/satsystems.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/satsystems.cpp
@@ -1431,6 +1431,18 @@ void Saturn::SystemsInternalTimestep(double simdt)
 	//FuelCellH2Manifold[0]->BoilAllAndSetTemp(315);		//Needs to be done using heat exchanger, heated to above 100F
 	//FuelCellH2Manifold[1]->BoilAllAndSetTemp(315);		//Needs to be done using heat exchanger, heated to above 100F
 	//FuelCellH2Manifold[2]->BoilAllAndSetTemp(315);		//Needs to be done using heat exchanger, heated to above 100F
+
+	if (this->systemsState < SATSYSTEMS_GSECONNECTED_1)
+	{
+		//need better solution
+
+		//keep the fuel cells from cooling prior to launch.
+		//this should be done by GSE, not this hacky piece of code
+
+		FuelCells[0]->SetTemp(475.0);
+		FuelCells[1]->SetTemp(475.0);
+		FuelCells[2]->SetTemp(475.0);
+	}
 }
 
 void Saturn::JoystickTimestep()
--- a/Orbitersdk/samples/ProjectApollo/src_csm/satsystems.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/satsystems.cpp
@@ -927,13 +927,6 @@ void Saturn::SystemsTimestep(double simt, double simdt, double mjd) {
 //double *pressequalFlow = (double*)Panelsdk.GetPointerByString("HYDRAULIC:FORWARDHATCHPIPE:FLOW");
 //sprintf(oapiDebugString(), "CSM Tunnel: %lf LM Tunnel: %lf TunnelFlow %lf EqFlow: %lf", (csmtunnelpipe->in->parent->space.Press)*PSI, (csmtunnelpipe->out->parent->space.Press)*PSI, (csmtunnelpipe->flow)*LBH, *pressequalFlow*LBH);
 
-	sprintf(oapiDebugString(), "FC1 %3.3fV/%3.3fA/%3.3fW FC2 %3.3fV/%3.3fA/%3.3fW FC3 %3.3fV/%3.3fA/%3.3fW",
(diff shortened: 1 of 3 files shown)


Thymo- (approved):
Looks good. All systems either have a custom minimum as listed in various documents or just use SP_MIN_DCVOLTAGE or SP_MIN_ACVOLTAGE (20V and 100V respectively). This might be a good starting point for working towards better power transient conditions as most systems just stop behaving right now. A good example would be the mission timer and computer for transient behaviour.

One thing I've noticed with the fuel cells is that there is no delay in flow increase with increasing loads. I'm not sure how quick that behaviour in the real fuel cells was but I can't believe it was instantaneous. Might be nice to look at as well.

n7275:
Good ideas. Converting to draft for now. I'll look into adding some transient reactant behavior.

n7275:
@Thymo-  @indy91 

Getting close turning this back into a non-draft PR.

One thing I shoud mention is that it will break any scenario with the line:
"<FCELL>  FUELCELL[1-3] 0 0.0000"

Would need to be changed to "<FCELL>  FUELCELL[1-3] 0 0.0000000000 0.0000000000"

I can go through and change all the mission scenarios, if you want. Or I can leave them, it's up to you.

Orbitersdk/samples/ProjectApollo/src_sys/PanelSDK/Internals/Esystems.cpp
```diff
@@ -387,11 +409,14 @@ void FCell::UpdateFlow(double dt)
 	}
 
 	//Idle power load to prevent the fuel cells from stopping and cooling too much (has to be fixed at some point)
-	if (power_load < 160.0)
-		power_load = 160.0;
+	//if (power_load < 160.0)
```
Thymo-, line 412: Personally I like to not leave commented out code as it takes away focus from the actual stuff going on. If it turns out it is needed again it can always be recovered from the git history. Do you agree with this?

Orbitersdk/samples/ProjectApollo/src_sys/PanelSDK/Internals/Esystems.cpp
```diff
@@ -445,30 +470,35 @@ void FCell::UpdateFlow(double dt)
...
-		}
+		Volts = 31.0; //we are trying to get 31.0V (1V per cell)
+
 		Reaction(dt, thrust);
 		running = 0;
 		Volts = Volts * min(1.0, reaction); //case of reaction problems :-)
 		if (reaction && Volts > 0.0)
 		{
-			Amperes = (power_load / Volts);
+		//make voltage (and current) drop by 5.2V over 1 day of normal impurity accumulation
```
Thymo-, line 489: Indent the comment and remove the extra whitespace at the start and end

Thymo- (changes requested):
Couple minor things but functionally looks good.

Orbitersdk/samples/ProjectApollo/src_sys/PanelSDK/Internals/Esystems.cpp
n7275: I agree. Bad practice from pre-git days.

Orbitersdk/samples/ProjectApollo/src_csm/satsystems.cpp
```diff
@@ -1424,6 +1424,18 @@ void Saturn::SystemsInternalTimestep(double simdt)
 	//FuelCellH2Manifold[0]->BoilAllAndSetTemp(315);		//Needs to be done using heat exchanger, heated to above 100F
 	//FuelCellH2Manifold[1]->BoilAllAndSetTemp(315);		//Needs to be done using heat exchanger, heated to above 100F
 	//FuelCellH2Manifold[2]->BoilAllAndSetTemp(315);		//Needs to be done using heat exchanger, heated to above 100F
+
+	if (this->systemsState < SATSYSTEMS_GSECONNECTED_1)
+	{
+		//need better solution
+
+		//keep the fuel cells from cooling prior to launch.
+		//this should be done by GSE, not this hacky piece of code
```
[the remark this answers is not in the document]
n7275: Agreed.

Orbitersdk/samples/ProjectApollo/src_sys/PanelSDK/Internals/Esystems.cpp
```diff
@@ -353,7 +375,7 @@ void FCell::Reaction(double dt, double thrust)
 	H20_waste->Flow(h2o_volume);
 
 	// TSCH clogging disabled
-    //Clogging(dt);
+    Clogging(dt);
```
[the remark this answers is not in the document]
n7275: changed spaces to tabs in last commit.

Thymo- (approved):

PR #563, opened 2020-12-08
reduce clogging by a factor of 3

n7275:
reduce clogging by a factor of 3 so we can make it to our purge interval without undervolt alarms

Will be replaced by a better solution soon


ThymoNL (approved):
Tested. Works well.

PR #603, opened 2021-04-19
LM ECS Stabilization Update

rcflyinghokie:
This update provides stabilization of the LM ECS system by better controlling glycol temperatures and pressures. Additionally, this provides more appropriate heat control in the LM cabin. These changes may break old scenarios.

LM ECS cold plate/rail systems implemented and heat loads directed to the plates instead of the glycol itself to more accurately represent the LM cooling system. This also prevents the secondary glycol loop temperature from climbing off scale high by having both loops attached the the appropriate plates.

Batteries now can generate heat based on current both during charge and discharge.  LM batteries initialized with temperatures and are connected to the LM cooling system.  CSM battery temperatures initialized to zero for now.

ECS/glycol loops in CSM now use glycol instead of water and have been initialized appropriately.

Nitrogen and CM cabin purge capability on pad and via waste stowage valve added.

This update also adds accumulators in a WIP state. They are not implemented in the spacecraft yet and need further testing.

More heat systems still need to be implemented so the LM runs cool.  Because of this, there is a known issues that suit temp control will actually likely cool the suit circuit if turned up to hot, this will be addressed in a future update.


Config/ProjectApollo/LEMSystems.cfg
```diff
@@ -119,73 +129,85 @@
...
+	<TANK> SUITCIRCUIT <0.5 0.5 -0.5> 30.0 0.001
 		CHM    0 0.02078149 0.02076071 8.857410234 # Calculated for volume at 0.01PSI and 0F
         VALVE IN   1 1.0								# from Suit Isol Valves
         VALVE OUT  1 1.0								# to SUITGASDIVERTER
 		VALVE OUT2  0 1.0								# Suit Circuit Relief to CABIN 
 	</TANK>
 	
-	<TANK> SUITGASDIVERTER <0.5 0.5 -0.5> 5.0 0.0001
+	<TANK> SUITGASDIVERTER <0.5 0.5 -0.5> 5.0 0.001
 	    CHM    0.005195374 0.005190178 2.214352559 # Calculated for volume at 0.01PSI and 0F
```
n7275, line 156: Should this be CHM **0** 0.005195374 0.005190178 2.214352559 # Calculated for volume at 0.01PSI and 0F

Or am I reading this wrong? There's a few others like this as well

rcflyinghokie: Good catch! While this doesn't hurt anything as the system fills up from the other sources, this was a copy paste error and will be fixed!

indy91 (approved):

PR #628, opened 2021-08-01
Rewrite telemetry socket cleanup

ThymoNL:
Before the PCM used some DLL hackery to close the telemetry sockets on sim exit. This was probably useful for something at some point but now it was only used for the SOCK page in the PA MFD. That page has been removed since it was no longer useful and was never intended to be released according to comments surrounding it.

Additionally when the LM PCM was implemented the CSM socket was overwritten by the LM socket causing the CSM socket not to be properly closed in scenarios containing a LM. This removes the calling across modules and cleans the sockets up in the PCM class instead.


indy91 (approved):

dseagrav:
> On Aug 1, 2021, at 11:00 AM, Thymo van Beers ***@***.***> wrote:
> Before the PCM used some DLL hackery to close the telemetry sockets on sim exit. This was probably useful for something at some point but now it was only used for the SOCK page in the PA MFD. That page has been removed since it was no longer useful and was never intended to be released according to comments surrounding it.
> 
This smells like the workaround for exiting back to the launcher not being the same as exiting Orbiter with ^Q.

Especially the part about it not knowing about the LM because the LM didn’t have telecom then.

The gist of the issue was that the DLL wouldn’t be unloaded if you launched another NASSP scenario and Windows didn’t have reusable sockets so the port would be blocked until Orbiter exited.

ThymoNL:
Ah that makes sense. At least nowadays Orbiter always destroys the vehicles so we now clean up the sockets in the PCM's destructor, this is always called even when quitting to the launchpad.

PR #633, opened 2021-08-08
Move pressure condition from LM hatch handles to LM hatches

rcflyinghokie:
Also adds a commented out abort discrete for Apollo 14


rcflyinghokie:
Also fixes LM suit isol valve actuator overrides so the manual disconnect can be performed with or without cb power

indy91 (approved):

PR #636, opened 2021-08-23
Convective radiators

n7275:
Attempt at a fix for EPS radiators cooling off on the pad to 0K. This seems to keep the ECS system semi-happy as well. I ran through the pre-launch on 7-11 (autochecklist, and time acceleration 10x-20x). *I believe* this should once again keep the LM's temps under control for Apollo 5, but I'm not good enough with the CBs to launch it yet. Temps are definitely *more* stable at 1x on the pad, but if my observations about time acceleration are the same as others, you'll break the AGC before you break EPS/ECS.

Without this fix, prelaunch is definitely usable. I only want to merge this if we all agree that it's distinctively better.


ThymoNL (approved):

ThymoNL (approved):
@rcflyinghokie has tested this without any problems. LGTM too.

PR #637, opened 2021-08-24
Fix a locking bug in the AGC code that caused EMEM corruption

ThymoNL:
When using time acceleration with multi-threading turned on the AGC timesteps are run from a separate thread and guarded by a mutex. When quickly switching between a high time acceleration factor and no time acceleration the thread sometimes hasn't finished before a new timestep is run. Because the no acceleration timesteps run from the main thread and do not check the mutex, undefined behavior occurs eventually causing the computer to trigger a restart from various conditions.

This commit always locks the mutex so that timesteps never execute at the same time. The language level is also set to C++14 so feel encouraged to use anything from that from now on.


indy91 (approved):

n7275 (approved):
Looks good.

PR #638, opened 2021-08-25
Fix LM steerable S-band antenna tracking under time compression

n7275:
also added slight improvement for jitter at high time compression using exp()


ThymoNL (approved):
Looks good to me. Let's wait merging this until mine is approved so they go in the same build.

indy91 (approved):

PR #645, opened 2021-09-11
Fix TEL TRUN 25° offset and Apollo 11 G&N PTC checklist

ThymoNL:
Two changes.

1. Set the correct SCT trunnion angle when the switch is set to 25°
2. Update Apollo 11s G&N PTC checklist. This checklist did not set the holdflag in the DAP. This needs to be done twice. First it is set to +0 to set the DAP to attitude hold. Then the steering command is loaded. After that the DAP is set to follow the steering command to start the roll maneuver by loading the holdflag with a negative value (70000).


rcflyinghokie (approved):
Checklist changes match the procedure and look correct.

PR #647, opened 2021-09-21
Variable center of gravity for the CSM

indy91:


Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
```diff
@@ -4342,7 +4353,7 @@ void Saturn::StageSix(double simt)
 
 {
 	//TBD: Use this function to shift the CG
```
ThymoNL: No longer TBD.

Orbitersdk/samples/ProjectApollo/src_lm/LEM.cpp
```diff
@@ -1972,7 +1972,13 @@ void LEM::SetStateEx(const void *status)
 void LEM::clbkSaveState (FILEHANDLE scn)
 
 {
-	SaveDefaultState (scn);
```
ThymoNL, line 1975: Long term idea. There's probably more stuff like this duplicated across NASSP vessels. We might want to make our own NASSP_VESSEL subclass from VESSEL4, add common code and have the LEM and CSM use that instead of VESSEL4.

jalexb88 (dismissed):

n7275 (dismissed):
No issues found.

Orbitersdk/samples/ProjectApollo/src_csm/saturnmesh.cpp
```diff
@@ -564,6 +564,12 @@ void Saturn::SetCSMStage ()
 	agc.SetInputChannelBit(030, GuidanceReferenceRelease, false);
 	agc.SetInputChannelBit(030, UllageThrust, false);
 
+	// Because all meshes are getting reloaded, we have to shift CG back to the center of the mesh, and then re-apply the offset CG on the next timestep
+	// Only necessary because of LET jettison function reloading all meshes.
+	ShiftCG(-currentCoG);
+	currentCoG = _V(0, 0, 0);
+	LastFuelWeight = 999999;
```
ThymoNL:
```suggestion
	LastFuelWeight = numeric_limits<double>::infinity();
```

Orbitersdk/samples/ProjectApollo/src_csm/saturn.h
```diff
@@ -1289,6 +1289,7 @@ class Saturn: public ProjectApolloConnectorVessel, public PanelSwitchListener {
 	// \ mass in kg, ro (distance from center of the bottom points), tdph (height of bottom points),
 	// \ height (of the top point), x_target (stiffness/damping factor, stable default is -0.5)
 	//
+	void ConfigTouchdownPoints();
 	void ConfigTouchdownPoints(double mass, double ro, double tdph, double height, double x_target = -0.5);
```
ThymoNL, line 1293: Is the idea to get rid of the paramaters completely? This is only called on staging and I see you check the stage variable in the new functions. If this function is kept I think the stage check should be done for all stages so the other function can be removed.

Orbitersdk/samples/ProjectApollo/src_csm/saturnmesh.cpp
ThymoNL: Same change can be applied here: https://github.com/orbiternassp/NASSP/blob/Orbiter2016/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp#L951

ThymoNL (changes requested):

Orbitersdk/samples/ProjectApollo/src_csm/saturn.h
indy91: ConfigTouchdownPoints gets called from the normal SetCSMStage function but now also from UpdateMassAndCoG, as touchdown points aren't automatically shifted with ShiftCG. So I just wanted to implement a common function that can be called from both places and I didn't want to have a duplicate of the fixed input parameters, so I made a version of ConfigTouchdownPoints without input parameters.

And I made it a switch/case because the function might be used by other stages in the future. Only by the CSM stage for now though.

Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
indy91: Will fix

Orbitersdk/samples/ProjectApollo/src_csm/saturnmesh.cpp
indy91: Then I also need to change it in the LEM as the same logic is used there.

Orbitersdk/samples/ProjectApollo/src_csm/saturn.h
ThymoNL: Okay! Whenever its logical the rest of the stages can be moved to the new function. Fine for now.

ThymoNL (approved):

PR #650, opened 2021-09-30
Change previous main and hover thruster groups to THGROUP_USER

indy91:
This update will disable the standard Orbiter controls for the main and hover engines, specifically the SPS, DPS, APS and all Saturn engines in the Saturn class. Previously it was possible to change the thruster level manually or a noisy joystick throttle lever could cause an early cutoff of the J-2 engine during TLI.


rcflyinghokie:
Since IRC is not working for me I will ask here...does this remove the ability to manually control throttle on the DPS?

indy91:
No, that was always custom code, although the Orbiter hover engine control was probably replicating the NASSP DPS engine controls. The only thing that is not working now is the main and hover engine thrust level displays on the empty internal view. The third view when you cycle through 2D and 3D panel with F8, that has two big MFDs.

rcflyinghokie (approved):
Sounds are different but nothing sounding "bad."

No other issues found.

PR #662, opened 2021-11-23
Rework ApolloGuidance interrupt handling

ThymoNL:
Streamlines things and properly handles interrupts during standby


n7275 (dismissed):
The code changes look good. This looks like a much cleaner and more streamlined solution. I tested standby mode and a few mission scenarios, along with both telemetry clients. As far as I can tell there are no issues.

indy91 (dismissed):

PR #669, opened 2022-01-06
Vesim update: DSKY functionality, CSM Abort and SPS related switches added

ggalfi:
Upon user request all three DSKYs (one in LM, two in CSM) buttons are now handled by Vesim. The idea is to use TouchPad as a virtual DSKY.
Some CSM functionality related to aborts and SPS-on-SCS burns also added to Vesim. These are:
- THC CCW&CW
- Direct Ullage pushbutton
- Thrust On pushbutton
- DV Thrust A&B switches
Please refer CSM - VKBsim Gladiator .cfg  for an example on how to bind a button-type Vesim input to an axis-type device, in this case the joystick throttle is used to select between the "Norm" and "Off" position of DV Thrust switches.


Orbitersdk/samples/ProjectApollo/src_sys/toggleswitch.cpp
```diff
@@ -417,6 +417,14 @@ bool TwoPositionSwitch::CheckMouseClickVC(int event, VECTOR3 &p)
 	return DoCheckMouseClickVC(event, p);
 }
 
+void TwoPositionSwitch::VesimSwitchTo(int newState)
```
ThymoNL, line 420: I think you can achieve the same result by using the existing TwoPositionSwitch::SetState function here.

ThymoNL (changes requested):
I think at least one instance of VesimSwitchTo could also work by using an existing function. The others don't really have an already existing function to do what you want.

It's probably a good idea to let SwitchTo play the click sound in the future. That'll come up again when support for XRSound is added.

Orbitersdk/samples/ProjectApollo/src_sys/toggleswitch.cpp
ggalfi: The idea was to have a general method for manipulating a switch, similar to CheckMouseClickVC and CheckMouseClick, but without the requirement to generate mouse click event or other unnecessary stuff. So what I wanted is a common function which processes a "deus ex machina" switching. Maybe my choice for the name wasn't the best, because such a method could be relevant outside of the Vesim bindings. Even the implementations of CheckMouseClickVC and CheckMouseClick would cleaner if they wouldn't be so redundant, and rather after the processing of the click events they would call the same method for the actual logic of the switch. Certainly it would be neat to extend this for all the buttons and switches, but I wanted to minimize the changes of the toggleswitch.cpp on first approximation.
On this specific TwoPositionSwitch::VesimSwitchTo(int newState): the TwoPositionSwitch::SetState does not check the attempt to set the switch to the same state as in it was originally. So it should be checked before each call of this function (there are four of them at the moment). It would result in redundant code, but additionally it seems to me that this type of logic is rather the part of the switch behaviour and as such, should be handled there.

rcflyinghokie:
@ggalfi will these changes impact the checklist MFD use of switching items? I see a lot of changes to DSKY inputs as well and, without testing it yet, I am just concerned the behavior could potentially impact it

ggalfi:
@rcflyinghokie 
> will these changes impact the checklist MFD use of switching items? I see a lot of changes to DSKY inputs as well and, without testing it yet, I am just concerned the behavior could potentially impact it

Regarding the checklist MFD, it don't have any impact, neither positive nor negative. In case of DSKY, the joystick inputs are not  sensed by the checklist - exactly as it happens when you are using the keyboard (Shift+Numpad keys). For the DV Thrust switches, I tested it in a few P40 situations and it cooperates nicely with the checklist - I guess in this case the MFD checks the status, not the actual switching event, but you are the expert of this. For DSKY I'd suggest to include the checklist management into the Dsky::...pressed() functions, or create some separate wrapper functions for that, and these could be used both by the original keyboard handling routine and also by Vesim. Other option is to query the display status instead of hunting for key presses - but it seems to me a pain in the ass when it comes to the practical details.

Orbitersdk/samples/ProjectApollo/src_sys/toggleswitch.cpp
ThymoNL: SwitchTo does check if you're trying to set the same state. If you are, it exits out and SetState won't play the sound. No issue there.
I agree with you that its nice to have this as a generic function. If you want, you can give them a nice generic name. In either case I''ll approve it.

ThymoNL (approved):

PR #678, opened 2022-01-16
Allow changing of checklist auto execution from MFD

ThymoNL:
With these changes the Checklist MFD gets a new button that allows you to enable or disable the automatic execution of the active checklist. Enabling or disabling the flag is only for the current simulation, the flag will revert to the state as configured in the PA Configurator after exiting to the launchpad.

There are two things that I would like some input on:
1. With this addition the MFD has an "auto complete" and an "auto execute" option. Should these be renamed, if yes, to what?
2. Should something be changed in regards to persisting a change from the MFD and reverting to what is in the PA Configurator?

New layout:
Bottom row
Menu

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ _$*
 *.ln
 *.hgtags
 core
+*.core
 
 # Project ignores (converted from .cvsignore files)
 !/Config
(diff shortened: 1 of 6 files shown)


rcflyinghokie (approved):
I think these descriptions work well. No problems I can see.

n7275:
I don't think I've commented on this feature over IRC, but I think being able to turn auto execution on and off from within the sim is a huge benefit.

PR #679, opened 2022-01-16
Virtual AGS commit: Incorporated a bug fix in yaAGS as suggested by r…

indy91:
…eader "spacex15".


ThymoNL:
I amended the commit to preserve the author information so it still shows Ron.

ThymoNL (approved):

PR #680, opened 2022-01-18
Prevent initial state of AEA engine on discrete, before AEA power up, to be interpreted as a valid engine on signal

indy91:
The check on the bits in the output register could be accomplished with less code.


ThymoNL (approved):
lgtm

jalexb88 (approved):

PR #681, opened 2022-01-19
CMVC & LMVC: Hatch controls were initialized with "AddSwitch" but eac…

jalexb88:
…h Hatch class already defines and operates its own hatch animations... All that is needed is to call the mouse click functions directly. This fixes the RR shaft animation erroneously being controlled by the LM hatch relief valves.


ThymoNL (approved):
Like the simpler solution. Nice work.

indy91 (approved):

PR #692, opened 2022-01-30
FDAI texture upgrade

jalexb88:
Lupus_Vulpes has authorized integration of new textures for the FDAI Ball. The textures are released under the GPL2 license. See here: https://www.orbiter-forum.com/threads/fdai-textures.40324/#post-590288


n7275 (dismissed):

rcflyinghokie:
Very pixelated and hard to read in 2D, looks great in the VC.

indy91 (changes requested):
Find solution for 2D panel.

ThymoNL (commented):
I would not be directly opposed to integrating these changes but as I understand someone is already working on new FDAI textures (by our request) that are higher resolution than this. To me it sounds like a lot of hassle to include these textures now if they will be replaced again within a short time span. Are the other textures expected to take a long time or do they textures provide a major improvement that is considered a must have to someone?

jalexb88:
> Find solution for 2D panel.

This should take care of it: https://github.com/orbiternassp/NASSP/pull/692/commits/bc007fd3023128144a9eb350ec9dd98d120037d2

Chrival:
Hi
I'm the guy who volunteered to make a new hi resolution FDAI texture.
Sorry for the delay but I already have a layout.
I opened a new thread at Orbiter forum
https://www.orbiter-forum.com/threads/new-fdai-tex-1800x900.40338/
Would it suit you ?

jalexb88:
@Chrival Yes by all means we would love your texture in NASSP! We have some competition for the FDAI textures it seems 😜 but both of you and Lupus_Vulpes works are very impressive. Chrival, since we came to you first I think it would be fair to include yours once they are ready.

Chrival:
Thanks Jalex. 
I know you're kidding but I don't want to compete with @Lupus. He's a fair man who gave me some useful advices this morning.
Besides I still need help:
Could someone give me the exact compression and format required for the .dds file ?
I guess it requires mipmap too. How many levels ?
Thanks in advance

Lupus:
I don't want to compete with you either, @Chrival , especially given the fact that I don't even know you 😂 Probably you wanted to at-mension someone else.

jalexb88:
@Chrival For the 2D panel: 1024x512 with some mipmap to help make it less grainy/pixelated maybe? For the 3D cockpit I think the 1800x900 you mentioned should be perfect with minimal mipmap to make it look crisp. Also, for the LM version we'll need one without the red areas.

Chrival:
I released the new textures, both 1024 and 1800:
https://www.orbiter-forum.com/threads/new-fdai-tex-1800x900.40338/post-590570
Unfortunately no mipmap but it nevertheless works.

Chrival:
I've done the LM FDAI too.
Here's the full set of textures. CM & LM
[FDAI.zip](https://github.com/orbiternassp/NASSP/files/8036833/FDAI.zip)

jalexb88:
I have updated this with the new textures by chrival, pull request is back open for review.

rcflyinghokie:
First impressions these look fantastic both in 2d and VC. My only observation (and this could be nit picky) is they seem maybe a little too tan? This could be fading from exposure on the museum panels. This also could be just an optical illusion being used to the white white background of them haha.

 I think its close but perhaps a touch too saturated? I would like to see others feedback on the "tanness" of these.
 
 No performance impact in 2D or VC and they are readable. I am excited to get some feedback on the color from others here as well before approving this PR.

maxq519:
My thoughts on the "tanness" thing are the same.  They probably weren’t true white initially, but probably more white than now… this stuff is now 50+ years old.

ThymoNL (dismissed):

ThymoNL:
Assigned this to myself. I'll squash the commits before being merged.

Chrival:
Sorry for the late answer, I was away from my PC this week-end.

@rcflyinghokie and @maxq519 I understand what you mean.
I did my best to match as close as possible the "true" colors, according to the pics that I have and taking into account that all these old stuff trend to turn yellow as time goes by. If I would have just picked up the color from these pics the FDAI would look like a true lemon ! I even corrected the balance of white, that's not hard since I assumed that the background of the rate scale are pure white (or pure light grey, it's quite the same).
IMHO it's rather authentic. Having flown old planes when I was young at the beginning of my career, I remembered it was something likeness.

But You may be right.
I've done a new one, less saturated. I took the opportunity to lighten a bit the white lines and digits (RGB 240 instead of 235).
I've only upgrade the CM FDAI 1800x900, for the moment. 
Let me know if you think it's better. In such a case I upgrade all the tex.
I'm thinking about darkenning the black lines and digits. They are RGB 24 and may be RGB 16 or 12 would be better. The whole tex would be more contrasted.

Edit : Attached File deleted

indy91:
Thank you @Chrival I will take a look as well! I think the FDAIs would at least look a little bit worn at launch, after all that testing and training they went through. I definitely like this updated one. If the others agree then it's probably this version that gets integrated into NASSP!

Chrival:
Ok so I'll prepare the new whole set and then you select.

jalexb88:
I just checked the texture, I think it looks great and blends in better with the rest of the VC textures.

rcflyinghokie:
Yeah this looks great, thanks for taking our feedback and adjusting so quickly! I will give it a full test in the CSM and LM tomorrow and add my approval after

Chrival:
And While I'im at it, what about the black lines and digits ? Do you want them a bit darker or is it ok ?

Chrival:
Here's the new set.
All black stuff are untouched.
Please report bugs or missing because each working image has around 90 layers. Not all has changed of course and I merged some to make the changes more quickly but I may have forgotten something.
Please disregard the old file CM.FDAI_Ball.LessSaturated1800.zip from a few post above since it's not exactly right. RGB 2**15**,207,1**96** instead of 2**16**,207, 1**95**. You can't catch it with the naked eye but **I** know it and it gives me hives and I deleted it btw.

[FDAIv1.zip](https://github.com/orbiternassp/NASSP/files/8058749/FDAIv1.zip)

rcflyinghokie:
waiting for the PR to include the new textures before approving

indy91 (approved):

jalexb88:
I have added the new set to the PR. Thanks @Chrival !

rcflyinghokie (approved):
Looks great here! Any improvements going forward can just be added in later PRs, I say this is good to publish.

Chrival:
Thanks to all
Just a rookie question : What is PR ?

rcflyinghokie:
> Thanks to all Just a rookie question : What is PR ?

Pull Request :)

ThymoNL (approved):

Issue #686, opened 2022-01-26
Update vessel names to missions with LM


rcflyinghokie:
Apollo 9 and 10 still use the AS-XXX designation for the CM instead of the name of the CM (Gumdrop, Charlie Brown.) These need to be changed at some point to match Apollo 11 and later where the CM has the actual CM name (Columbia etc)

PR #699, opened 2022-02-03
CSM_name_fix

maxq519:
Fixed the Apollo 9 and 10 scenarios so the CSM uses the correct name instead of AS-5XX numbers


rcflyinghokie:
Charlie Brown is not hyphenated.

rcflyinghokie:
Closes #686

rcflyinghokie:
Need to verify with @indy91 that these name changes won't impact MCC/RTCC mission specific data loading

ThymoNL (changes requested):
Thanks for fixing this!
It looks like all lines were changed to use Windows line endings. Your editor should have a setting to change to Unix line endings (\n). Please check this.
Can you change your commit message to say "Fixed the Apollo 9 and 10 scenarios so the CSM uses the correct name instead of AS-5XX numbers" instead of "Add files via upload"?
After you've fixed that, can you recreate the PR so that the source branch points to your CSM_name_fix branch directly instead of Orbiter2016? That removes the extra merge commit.

maxq519:
Ok, so I am totally clueless when it comes to GitHub...
Is there somewhere I can read up on how to do these things?

As far as the line endings thing, I can easily fix that.

ThymoNL:
I always recommend this as a quick intro: http://rogerdudler.github.io/git-guide/
That shows you how to work with the git client.
When you create a PR you pick the source branch from the compare window. This should be CSM_name_fix

maxq519:
So do you want me to delete this PR and open a new one?

ThymoNL:
I don't think you can change it after you already opened it so yes.

PR #700, opened 2022-02-03
CSM name fix

### LINKED ISSUE #686: Update vessel names to missions with LM

maxq519:
Fixed the Apollo 9 and 10 scenarios so the CSM uses the correct name instead of AS-5XX numbers


ThymoNL:
Closes #686

rcflyinghokie:
Need to verify with @indy91 that these name changes won't impact MCC/RTCC mission specific data loading

indy91:
It seems like for Charlie Brown we already had some code in the PAMFD and the RTCC MFD to detect the vehicle and/or mission number. And for Gumdrop the PAMFD has that, too, just for the RTCC MFD I failed to add it. So for the RTCC MFD to detect that Gumdrop means it is the Apollo 9 mission there is a one line code change required.

In the file ARCore.cpp

`else if (strcmp(v->GetName(), "AS-504") == 0 || strcmp(v->GetName(), "Spider") == 0)`

to

`else if (strcmp(v->GetName(), "AS-504") == 0 || strcmp(v->GetName(), "Gumdrop") == 0 || strcmp(v->GetName(), "Spider") == 0)`

I think in this way even old scenarios will still work where the name is still AS-504 and AS-505. I'll think about any other places where the name change could be an issue, but it might be that this is the only place other than the MCC part of the scenarios.

indy91 (approved):
I think this should do it.

rcflyinghokie (approved):

PR #701, opened 2022-02-05
Change entry guidance padloads to their actually used values on Apoll…

indy91:
…o 14-17


ThymoNL (approved):
It looks like this pulls in the aero stuff again. That is a side effect of the maintenance I just did to Orbiter2016, so ignore that.

PR #708, opened 2022-02-11
Update Saved Flightplans

rcflyinghokie:
This updates the flight plans stored in the NASSP directories to replace incomplete ones and combine multiple pdfs. Also moves old word flight plans.

This should stay in draft until we decide if we will ship these with NASSP or a dedicated repository similar to padload worksheets.


rcflyinghokie:
I am closing this for now until we decide how to handle this going forward with documents we have in our repo

PR #710, opened 2022-02-13
MCC: Fix Apollo 10 Block Data 1 calculation

indy91:


ThymoNL (changes requested):
Your branch has diverged. This causes each PR you make to include all commits from the point your branch diverged. This problem will only grow over time.
You need to rebase onto the current Orbiter2016 to fix this.

ThymoNL (approved):

PR #711, opened 2022-02-13
Consume Ctrl-D undock key to prevent invalid probe state

ThymoNL:
It would be nicer if Orbiter exported the current keymap through the API so the Undock key itself can be consumed instead of the default keybind for it.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
@@ -3371,6 +3371,12 @@ int Saturn::clbkConsumeBufferedKey(DWORD key, bool down, char *kstate) {
 		return 0;
 	}
 	if (KEYMOD_CONTROL(kstate)) {
+		switch (key) {
+			case OAPI_KEY_D:
+				// Orbiter undocking messes with our undocking system. We consume the keybind here to block it.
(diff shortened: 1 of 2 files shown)


n7275 (approved):
I think this probably the only solution without changing Orbiter/oapi. Should prevent any accidental undocking, caused by orbiter controls. At the moment you cant even change Orbiter's defaults, so this should be safe and will work the same for everyone.

rcflyinghokie:
No issues here

jalexb88:
I've had that happen a few times myself in the CSM during TLC, unknowingly hit ctrl-D, after a while look outside and the LM is nowhere to be found. I think this is a welcome change.

ThymoNL:
> I think this probably the only solution without changing Orbiter/oapi. Should prevent any accidental undocking, caused by orbiter controls. At the moment you cant even change Orbiter's defaults, so this should be safe and will work the same for everyone.

Oh you can totally change them by editing keymap.cfg. This is backed up by Orbiter's documentation. Orbiter tries to map any keybind to an action configured in the keymap if the keypress is not consumed by something else first.
See here: https://github.com/orbitersim/orbiter/blob/main/Src/Orbiter/Keymap.cpp
and here: https://github.com/orbitersim/orbiter/blob/c062423dc8c725c0d403c96a8a268e9a7d0ffd9e/Src/Orbiter/Vessel.cpp#L3595

PR #722, opened 2022-02-25
Correct Apollo 8-11 Fuel Cell pre-launch configuration

CaptainSwag101:
Missions up through Apollo 12 used FC 1 & 2 to power MNA, and FC 2 & 3 to power MNB, while Apollo 13 and later missions used only FC 3 to power MNB. Apollo 7 has been left alone (aside from the pre-launch checklist) because that mission performed several FC configuration changes in flight that I am not educated about.

~~Please note that this will mean the pre-saved mission scenarios will not have the proper H2 and O2 cryo tank quantities as you get further into the mission, as the increased flow to FC2 is not propagated throughout the duration of the mission prior to where the scenario save was made. However it is my understanding that these scenarios will need to be remade anyway due to many other changes made over time, so I do not believe this is a serious issue.~~

~~Additionally, it should be noted that at the time of committing, the "just before launch" scenarios for Apollo 9 and 10 trigger a pad abort upon loading in, but this is not an issue caused by my changes. I will be creating an issue report for this problem shortly.~~


CaptainSwag101:
It has been brought to my attention that this change would actually not impact the H2 and O2 cryo tank quantities at all, since the total power load is the same, just distributed differently. Disregard my note on that topic.

ThymoNL (approved):
Looks good to me. I appreciate the detailed commit message. Thanks!

PR #723, opened 2022-02-26
Improve RTCC calculations for SM RCS deorbit maneuvers

indy91:
-Earth Orbit Entry PAD can simulate deorbit maneuvers that are not using the SPS and non-standard reentry profiles
-Slight adjustment of RTCC system parameters for the SM RCS, also very small adjustment of the actual thrust values of the SM RCS thrusters
-Renaming a lot of retrofire input variables to make it clearer to which maneuver they belong

More work needed to improve iteration behavior of fixed DV SM RCS deorbit burns.


Orbitersdk/samples/ProjectApollo/src_launch/rtcc.h
```diff
@@ -3949,28 +3952,51 @@ class RTCC {
 	RetrofireDisplayParametersTable RZRFDP;
 	SpacecraftSettingTable RZDBSC1;
 
+	struct ReentryOutputTable
```
ThymoNL, line 3955: Empty struct?

ThymoNL (changes requested):

ThymoNL (approved):

PR #727, opened 2022-02-28
fix slow water dump for missions started from prelaunch

n7275:
the missions an be fixed too, but I want to write a script to do it, rather than doing it manually.


ThymoNL (approved):

ThymoNL:
Looks good to me. Do you want to add the fixed scenario's to this PR or separately?

n7275:
I will do them separately. It might take me a few days and I wanted to get this one out of the way.

PR #732, opened 2022-03-03
Add label files for Apollo Earth Landmarks

indy91:


ThymoNL (dismissed):

n7275 (dismissed):

ThymoNL:
I like this solution. Nice to see that Orbiter's file can be left untouched.
Can you squash your commits into one so it becomes a single change?
If you don't know how to: https://stackoverflow.com/questions/12522565/how-can-i-combine-two-commits-into-one-commit

indy91:
There are still problems, likely a bug in Orbiter.

rcflyinghokie (dismissed):
Tested on a few LEO missions, markers in each hemisphere are visible and seem to be in the correct location visually.

rcflyinghokie (approved):

ThymoNL (approved):

PR #739, opened 2022-03-11
Add "wait for XX" for MCC decision updates

rcflyinghokie:
Add time checks to Apollo 11 checklist to better facilitate off flight plan times since the uplinks are relative to other maneuvers and not necessarily the flight plan times.


indy91 (approved):

dseagrav:
If the time of thing being waited for is available as a (C++) variable, it should be able to show up in the checklist using the checklist item variable substitution hack that I added.

Issue #641, opened 2021-08-30
Cabin ambience volume


ThymoNL:
Some people have indicated that they experience the cabin sounds to be too loud for comfort, especially when using the simulator for extended periods.  Currently the volume of things like the cabin fans and suit compressors are fixed.

Too alleviate some headaches and to make the experience more enjoyable a volume control should be added to the PA configurator allowing users to adjust the equipment volume to a comfortable level.

CaptainSwag101:
I would like to bump this issue because I agree that especially in the CM, the ambient noise is rather unpleasant when systems are configured per usual mission parameters (Suit Compressor on, and both Cabin Fans active).

First and foremost, I think the suit compressor sound should be attenuated if possible. Since the compressor is likely behind several panels deeper within the CM, the high-pitched whine it makes should be less audible than the sound of the circulating air. I don't know if it's possible to adjust the parameters of OrbiterSound to apply something like a low-pass filter, but if not, I have personally achieved a similar effect by putting the original WAV sound file in Audacity, applying a 1000 Hz low-pass filter with -12 dB rolloff, and then amplifying by 2 dB. This results in the lower-pitched sound of rushing air to remain mostly the same volume, while muffling the high-pitched whine of the compressor to a much more tolerable level, while not eliminating it entirely. This on its own makes the cabin ambience far less grating on the ears, in my subjective opinion. To take things further, the cabin fan volume could optionally be reduced slightly (or the astronaut could use only one cabin fan, but that would deviate from normal mission parameters), but the suit compressor sound is probably the primary culprit for uncomfortable listening.

For your consideration, here is a copy of the altered suit compressor sound effect WAV file.
[suitcompr_muffled.zip](https://github.com/orbiternassp/NASSP/files/8234916/suitcompr_muffled.zip)

n7275:
can this be closed?

PR #742, opened 2022-03-11
Cabin ambience adjustments

CaptainSwag101:
Based on Issue #641 I have opted to make a few alterations to the CM Suit Compressor sound to both alleviate the harshness of the sound during prolonged simulator sessions, and to potentially make the sound more accurate to what was heard in the cockpit in reality. Since the suit compressor unit was presumably not located directly in the cabin, the paneling and other obstructions would attenuate the sound, in particular the high-pitched whine of the compressor would be reduced.

To accomplish this effect, I took the original SFX file and used Audacity to apply a 1000 Hz low-pass filter with -12 dB rolloff, then amplified the whole audio by 2 dB. Afterwards I applied a reverb effect, whose settings are pictured here:
suitcompressor_reverb_settings

Overall, I think this helps make cabin ambience much more pleasant for long durations, though other changes could perhaps be incorporated such as tweaking the sound of the cabin fans as well. However I believe the primary culprit for ear strain is this suit compressor noise, and this PR should hopefully help make things more tolerable.


ThymoNL (approved):
Sounds good to me. I'll let others have a listen too.

indy91 (approved):
Definitely an improvement.

rcflyinghokie (approved):
Sounds good here!

PR #743, opened 2022-03-12
Apollo 7 MCC update

indy91:


indy91:
I have tested all the updated scenarios and they work fine. This update will break old scenarios, but it makes the Apollo 7 MCC more flexible for additional updates and calculations in the future. I will make a post in the release thread when this gets merged.

In addition to restructuring the mission states I added two update, a S-IVB state vector uplink and the PAD for the retro attitude orientation test. I had these two updates in my branch for Apollo 7 with realistic aerodynamics for a long time and they are well tested.

ThymoNL (approved):

rcflyinghokie (approved):

PR #751, opened 2022-03-21
Apollo 9 checklist add antenna transition time text & Apollo 10 checklist fixes through LM SBD Checkout

rcflyinghokie:


rcflyinghokie:
Need to investigate which checklist fix causes Apollo 9 LM scns to CTD

rcflyinghokie:
CTD fixed, Apollo 9 scns no longer impacted by checklist issues.

ThymoNL:
Do you have a little bit more background info?
If a scenario can cause a CTD that sounds like something that should be addressed.

rcflyinghokie:
Unfortunately I am not sure, I reverted back to the original checklist from the main repo and it no longer CTD. 

It only happened on Apollo 9 when a LM was present, and it threw this exception:

Exception thrown at 0x3AF485F2 (LEM.dll) in orbiter.exe: 0xC0000005: Access violation reading location 0x5487BE16.

So all I did was revert and test the changes one by one, I think I may have entered something incorrectly or copy paste issue in the wrong column not sure, but in any case all Apollo 9 scns open properly.

The two changes were changing the flight plan start time and adding yellow text before the LR spurious test, so I think I might have had some sort of formatting error in one of those.

indy91:
The scenarios save the position in the checklist, but when the checklist itself changes in a way where that position doesn't exist anymore a CTD happens. I've tried figuring it out once, but the checklist code is quite complex. And the best we could do is prevent a CTD by "resetting" the position in the checklist. And for the mission scenarios that isn't that much better than a CTD. But it's an issue that needs to be addressed in the long term.

indy91 (approved):

PR #756, opened 2022-03-25
MCC: Correct formats of Apollo 7 and 9 Manuever PADs

indy91:


Orbitersdk/samples/ProjectApollo/src_launch/mcc.cpp
```diff
@@ -2412,11 +2412,66 @@ void MCC::drawPad(bool writetofile){
 		{
 			int hh, mm;
 			double ss;
+			char buffer2[1024];
```
ThymoNL: These names are not descriptive.
Looking at your code buffer2 is some kind of intermediate that you use to construct the current line and buffer 3 is the string of the complete PAD.

Orbitersdk/samples/ProjectApollo/src_launch/mcc.cpp
```diff
@@ -2412,11 +2412,66 @@ void MCC::drawPad(bool writetofile){
...
+			{
+				buffer3 = "MANEUVER UPDATE (P30)\n";
+			}
+			else
+			{
+				buffer3 = "MANEUVER\n";
+			}
+
+			SStoHHMMSS(form->GETI, hh, mm, ss);
+			sprintf_s(buffer2, "%s PURPOSE\n%+06d HRS GETI N33\n%+06d MIN\n%+07.2f SEC\n%+07.1f DVX\n%+07.1f DVY\n%+07.1f DVZ\n", form->purpose, hh, mm, ss, form->dV.x, form->dV.y, form->dV.z);
```
ThymoNL: sprintf_s is a non-portable Windows only fuction.
We want NASSP to be portable so it can easily be compiled for Linux when Orbiter starts supporting it, so we want to avoid introducing new portability issues whenever we can.

You can use snprintf which does the same but in a portable way, you just need to supply a size yourself.

ThymoNL (changes requested):

Orbitersdk/samples/ProjectApollo/src_launch/mcc.cpp
indy91: Yeah I need to get in some consistency with these functions, we are using them a lot in the MCC and RTCC MFD. Originally I was only using sprintf, but sprintf_s had the advantage of causing a guaranteed CTD when there was a bug with the code (character array size not larger enough etc). I can use snprintf from now on and slowly convert all sprintf and sprintf_s if that is the right function for the job.

Orbitersdk/samples/ProjectApollo/src_launch/mcc.cpp
indy91: See, you figured it out easily. Descriptive enough. If you suggest some names I'll change it, but I don't think it's super necessary.

ThymoNL: That is not really what I meant...
I was hoping my explanation of what I think you are trying was enough of a hint for a proper name.

In order to understand what `buffer2` means I have to read a whole section to understand its purpose.
If they would've been named something like `padLine` or `currentLine` I wouldn't need to spend 20 seconds reading ahead just to understand what you are even trying to do.

indy91: Well "line" is more confusing as even the updated code doesn't handle the PAD line by line. I will use tempString and fullString, I think that gets across what it does enough.

n7275 (approved):

ThymoNL (approved):

PR #757, opened 2022-03-27
Attempt to fix building with Visual Studio 2022

CaptainSwag101:
So far, it seems that setting the S1B project to depend on the PanelSDK project in the solution properties, fixes the one build issue that prevents things from "just working".  An alternate solution is to change the S1B project's linker directory to explicitly point at the full absolute path of the PanelSDK build output directory rather than the relative/implicit path that it currently uses, but this is far from ideal. The aim is to solve this build issue using a method that does not require each user to alter the build configurations to suit their local machine, as such changes are not portable, and cannot and should not be brought back to the upstream repo.

However I can only test this on my own machine, which isn't good enough. For reference, my setup is using VS 2022 Community with the latest Windows 10 SDK (though I am using Windows 11) and with only the latest VC++ toolchain, v143. Fingers crossed it works for everyone, C++ is weird sometimes.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/Build/VC2017/ASTP.vcxproj
+++ b/Orbitersdk/samples/ProjectApollo/Build/VC2017/ASTP.vcxproj
@@ -12,20 +12,20 @@
   </ItemGroup>
   <PropertyGroup Label="Globals">
     <ProjectGuid>{BAE34D87-5EAE-491C-AE6D-73B6B1E487D2}</ProjectGuid>
-    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseOfMfc>false</UseOfMfc>
     <CharacterSet>MultiByte</CharacterSet>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v143</PlatformToolset>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseOfMfc>false</UseOfMfc>
     <CharacterSet>MultiByte</CharacterSet>
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v143</PlatformToolset>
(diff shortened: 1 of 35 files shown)


CaptainSwag101:
I realize now that these changes will almost definitely break compilation on VS 2017 and 2019 since they upgrade the toolchain. I will actually be going back and readjust to see if I can keep the project working with the existing v141 toolchain. Stand by.

rcflyinghokie:
If you still want to work this one just make it a draft

dseagrav:
> On Mar 27, 2022, at 11:56 AM, James Pelster ***@***.***> wrote:
> So far, it seems that setting the S1B project to depend on the PanelSDK project in the solution properties, fixes the one build issue that prevents things from "just working". An alternate solution is to change the S1B project's linker directory to explicitly point at the full absolute path of the PanelSDK build output directory rather than the relative/implicit path that it currently uses, but this is far from ideal. The aim is to solve this build issue using a method that does not require each user to alter the build configurations to suit their local machine, as such changes are not portable, and cannot and should not be brought back to the upstream repo.
> 
Requiring the build to be edited for the machine it is executed on will hard break the auto-build process as well, if that’s still used.

CaptainSwag101:
Agreed, my goal is to avoid anything like that, I want things to be as painless as possible for autobuild systems and contributors alike. The solution as I see it, is to upgrade the Windows 10 SDK version for each project, which along with setting the S1B project dependency will allow things to build properly on VS 2022 while using the old v141 toolchain, which is still available in the installer. The only reason I need to upgrade the Win10 SDK is because the version currently used by the project is no longer supported in VS 2022 and cannot be installed via its installer.

CaptainSwag101:
Curiously, if I revert all these changes and manually install the expected version of the Windows 10 SDK that the project already wants, I get over 400 fatal errors in my build, but it... actually builds just fine??? Why is C++ so horribly messed up, seriously...

spacex15:
I would add the following lines to your .vcxproj files

    <WindowsTargetPlatformVersion Condition="'$(VisualStudioVersion)' == '15.0'">10.0.17763.0</WindowsTargetPlatformVersion>
    <WindowsTargetPlatformVersion Condition="'$(VisualStudioVersion)' == '16.0'">10.0.18362.0</WindowsTargetPlatformVersion>
    <WindowsTargetPlatformVersion Condition="'$(VisualStudioVersion)' == '17.0'">10.0.19041.0</WindowsTargetPlatformVersion>    
instead of what was already there (WindowsTargetPlatformVersion)

and

    <PlatformToolset Condition="'$(VisualStudioVersion)' == '15.0'">v141</PlatformToolset>
    <PlatformToolset Condition="'$(VisualStudioVersion)' == '16.0'">v142</PlatformToolset>
    <PlatformToolset Condition="'$(VisualStudioVersion)' == '17.0'">v143</PlatformToolset>

instead of  what was already there (PlatformToolset )

and you will have a multi compatible project (2017,2019,2022)

ThymoNL:
I think a more durable solution would be to convert NASSP to a CMake project and let it figure out the correct build environment. If this is still needed in the future this PR can be reopened. Closing for now...

PR #759, opened 2022-03-27
Adjust electroluminescent display color and digit design

CaptainSwag101:
Recent analysis of a real electroluminescent DSKY display has determined the peak wavelength it emits to be around 514 nm, which appears to be roughly a lime-green color. However, this color is not accurately represented in the sRGB colorspace, and the light produced by these displays is extremely wide-band compared to the light emitted by, say, an LED, meaning it can look different in various ambient lighting conditions. Eyewitness accounts also indicate that the display often appears somewhat bluish in many lighting conditions.

Therefore, I have tried to strike a nice balance between the current color (a mostly cyan color) and the direct sRGB representation of the appropriate wavelength (bright lime-green, much like the color used in NASSP 7.0). I took the current cyan textures and hue-shifted them in the "paint.net" image editing tool by a factor of -10 (negative ten). This results in an RGB value of #10FF9D, which roughly corresponds to a wavelength of 499/500 nm. This is not quite the target wavelength, but I feel it may be a good compromise between the two different shades that have been used in the past, and is hopefully a good "average" of the color people tend to see when looking at an original display in person.

I expect there may or may not be controversy and debate around this change, at the end of the day I'm just trying to tweak things to get them slightly more accurate than they were before, and I don't believe the previous cyan shade was quite right.

Since it may be difficult to appreciate the differences by simply comparing texture files, I have provided several screenshots of the new display colors and how they appear on both the DSKY and other EL displays in the CSM and LM, in both 2D view and the 3D virtual cockpit.

orbiter_2022-03-27_12-06-45
orbiter_2022-03-27_12-07-01
orbiter_2022-03-27_12-07-20
orbiter_2022-03-27_12-09-39
orbiter_2022-03-27_12-09-51


CaptainSwag101:
I would like to request that while people weigh in on the new colors, I ask that we refrain from merging this PR until I can take care of a few things. Firstly, I want to verify that the VC digits are actually the same color, as some folks have mentioned that it looks like they appear less vibrant than those on the 2D panel, though this could also be due to the lighting of the VC itself being dimmer.

I am also working on recreating these EL digits in vector form, based on the original NASA specs. I hope to then turn them into bitmaps so that we have a more accurate representation of the digits, assuming they look good when scaled down to that size.

CaptainSwag101:
As a further part of this PR, I have also recreated the electroluminescent digit bitmaps using the original NASA design schematics as closely as possible. I have adjusted the bitmaps from the original full-resolution source image in order to make them work properly with NASSP's method of rendering, but it should be noted that the digits are slightly less wide and their skew angle is slightly less than spec, due to the very slight horizontal squeeze that I needed to apply in order to make digit widths match the old digit textures. In the future it would be nice to rework the texture display system in such a way that those modifications are no longer necessary, but for the time being I am willing to compromise because the new digits are more accurate to their shape on the real spacecraft either way. I should also note, the digit colors have been slightly tweaked to #00FF96.

Here are some screenshots using the new digits:

orbiter_2022-03-27_16-59-42
orbiter_2022-03-27_17-00-22
orbiter_2022-03-27_17-07-12

And, for those who care, here is the original-resolution digit set (keep in mind the scaled bitmaps were also tweaked slightly in size and shape after the fact):
NASSP Digit redesign

CaptainSwag101:
One more last-minute change to make: I need to fix the + and - segments for the DSKY on the 3D VC, I forgot to correct that texture after fixing the 2D one.

CaptainSwag101:
All right, everything should be good to review now; I'll make changes as needed but as far as I'm concerned this should be okay.

n7275 (dismissed):
This looks good to me, but I want other's input too.

CaptainSwag101:
Absolutely, I'd really appreciate some critical eyes on these changes. If this is gonna be "the look" of NASSP 8.0 and onward then I want things to be about as perfect as can be.

ThymoNL:
> due to the very slight horizontal squeeze that I needed to apply in order to make digit widths match the old digit textures

I'm sure this can be done. Would just need to extend the texture slightly and update the coordinates each digit is at.

ThymoNL (dismissed):

rcflyinghokie:
Has this been applied to all EL displays?
CSM:
MDC DSKY
MDC Mission Timer
EMS Display
LEB DSKY
LEB Mission Timer

LM:
DSKY
DEDA
Mission Timer
Event Timer
Propellant Quantity Display (x2)
Helium Display

CaptainSwag101:
Yes, all of these displays are currently updated with the new font/color, including in the 3D virtual cockpit. However, I am in agreement with Thymo that if we can rework the display-painting code so that no bitmap-tweaking is necessary, that would be ideal. It would make the font rendering even more accurate by not squashing the digits horizontally just to make them fit the constraints of the old ones.

CaptainSwag101:
Another update (sorry, I just keep finding minor discrepancies with my digits versus the original specs), I am updating the plus and minus digits once again to make them larger, looking at the documents again and source footage shows they are actually quite large, and this also helps improve visibility when they will inevitably be scaled down to the final bitmaps.

CaptainSwag101:
Okay, I have spent several hours going over everything with a fine-toothed comb so to speak, and I think this should be at last ready for review again. I have refactored the EL drawing code so that I can use my new digits without squishing them, and the new code should hopefully be easier to maintain and configure in the event that the digit bitmaps change again, with local constants for digit dimensions, etc.

I have also re-rendered all of the digit bitmaps for 2D and 3D views, previously due to the horizontal squeeze they did not have perfectly consistent anti-aliasing or pixel alignment, resulting in slight variations in digit position and shape. The new renders are pixel-perfect similar to the old digit design and look much more appealing as a result.

Please let me know if anything doesn't look right or needs to be adjusted, but as far as I know every EL display in both 2D and 3D should be fully functional and correct.

n7275 (approved):

indy91 (approved):
Colors look good. The new digits look a bit strange to me, but I am sure they are correct and I just need to get used to them. Of course these digits would only apply to the DSKY, the other digital displays had other manufacturers and would likely look different. But we won't have the exact specifications for those. It's something we need to revisit some time though.

ThymoNL (approved):

rcflyinghokie (approved):
Certainly will take some getting used to, and probably need iterations along the way, but I see no reason not to press on.

Cue the complaints on the forums :P

ThymoNL:
@CaptainSwag101 Wanna do the honors of writing up a little update for the forum?

CaptainSwag101:
> @CaptainSwag101 Wanna do the honors of writing up a little update for the forum?

Sure thing, I'm at work currently but I'd be happy to once I get home in a few hours.
It seems as though the NASSP 8.0 thread is locked except for dev team members, perhaps?

PR #764, opened 2022-04-01
RTCC: Disabling access to LDPP mode 6 broke all the other modes, too.…

indy91:
… This makes sure only mode 6 is disabled.


indy91:
f8ebdcb82c243c4a233004eefea953c0d7c1ba10 is what broke it.

ThymoNL (approved):

PR #768, opened 2022-04-08
Add improved scenario descriptions

maxq519:
I added html scenario descriptions for Apollo 7-11.


indy91:
Looks great! Where are the texts describing each mission from?

maxq519:
I got the text from https://history.nasa.gov/
I put credits info in the file “/html/Scenarios/NASSP/credits.txt”.

ThymoNL (changes requested):
Can you check that the margins are consistent in each file?

image
image
image

maxq519:
Will do.  I didn’t notice that!

ThymoNL (dismissed):
Looks good. I'll wait to hear if Ryan has anything else.
Thanks. :-)

CaptainSwag101:
I have only checked the Apollo 7 description so far, but I have noticed small issues as shown below with lines being spaced unevenly, or cut off unexpectedly in the blank area between lines.

orbiter_GyN4mcpWJT
orbiter_YLJCMaaStS

maxq519:
> I have only checked the Apollo 7 description so far, but I have noticed small issues as shown below with lines being spaced unevenly, or cut off unexpectedly in the blank area between lines.
> 
> orbiter_GyN4mcpWJT orbiter_YLJCMaaStS

Yeah, I noticed that too.  I think it happens on lines where there is a subscript or superscript… Not really sure what I can do about it, except getting rid of the superscripts/subscripts.

rcflyinghokie:
Personally I think this is too much detail for a mission description in a scenario file. I think it needs to be shortened and not have a novel.

Additionally, the title is a bit large and I really dislike "The Fifth Mission" etc as it really serves no purpose.

maxq519:
> Personally I think this is too much detail for a mission description in a scenario file. I think it needs to be shortened and not have a novel.
> 
> Additionally, the title is a bit large and I really dislike "The Fifth Mission" etc as it really serves no purpose.

You got it.  I kinda agree on the “The Fifth Mission” thing, it was ok at first, but each time I look at it, I like it a little less.

I will shorten the write-ups, but how about the pictures?  Should they stay or go?

rcflyinghokie:
> > Personally I think this is too much detail for a mission description in a scenario file. I think it needs to be shortened and not have a novel.
> > Additionally, the title is a bit large and I really dislike "The Fifth Mission" etc as it really serves no purpose.
> 
> You got it. I kinda agree on the “The Fifth Mission” thing, it was ok at first, but each time I look at it, I like it a little less.
> 
> I will shorten the write-ups, but how about the pictures? Should they stay or go?

I would say they are unnecessary. People will spend very little time on the description as it is, so saving space and lines etc is probably prudent in this case.

maxq519:
True, but I know at least in my case, if there are pictures in the description it might make me spend a little more time reading it… the Project Mercury X add on has pictures in the descriptions and it really adds something to the effect.  If you look at the level of detail in my NASSP descriptions versus the Project Mercury ones, it is about the same.

CaptainSwag101:
I think one or two images per scenario might be worthwhile, maybe a picture of the crew and of a prominent moment in the mission, perhaps? I think that ought to be a good compromise.

indy91:
I agree with @CaptainSwag101, nothing too elaborate, but a small number of pictures and a general mission description would be good to have.

ThymoNL (approved):

PR #772, opened 2022-04-25
System worker threads

n7275:
This is very much not-mergeable. It is mostly here as a placeholder. If I can't optimize it to the point where it is consistently better than the linked-list update version, than I will close the PR.

Two big issues to solve:
* It's slower than the old version it's intended to improve upon, I believe due to the std::vector copy operation in StartWork.
* It causes AGC restarts on scenario load and I don't know why. ~~Fixed~~

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_sys/PanelSDK/SPSDK_WorkerThread.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_sys/PanelSDK/SPSDK_WorkerThread.cpp
@@ -36,6 +36,7 @@ ThreadPool::ThreadPool()
 {
 	dt = 0;
 	terminate = false;
+	queuePosition = 0;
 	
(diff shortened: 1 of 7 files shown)


Orbitersdk/samples/ProjectApollo/src_sys/PanelSDK/SPSDK_WorkerThread.cpp
```diff
@@ -0,0 +1,121 @@
...
+				if (++idleThreads > numThreads) { idleThreads = numThreads; }
+				cvWork.notify_all();
+				cv.wait(RunningLock);
+			}
+		}
+
+		{
+			std::unique_lock<std::mutex> lock(readQueueLock);
+			if (!queue.empty()) {
+				Job = queue.back();
```
n7275: This is slow

PR #774, opened 2022-05-04
MFD fixes and switch failure function

indy91:


n7275 (dismissed):

Orbitersdk/samples/ProjectApollo/src_mfd/ProjectApolloMFD.h
```diff
@@ -123,6 +123,14 @@ class ProjectApolloMFD: public MFD {
 
 protected:
 
+	//Utility functions
+	bool IsSaturnIB(VESSEL *v);
```
ThymoNL: I don't think this is a scalable solution. This now requires a separate function for each vessel to be checked. In the future if one wants to check if the VESSEL is a LRV, Skylab, Wet workshop, etc, these would also need to be added.
It would be better to have a single IsVessel and give it an enum input with the vessel to be checked.

Also this can better be moved to its own file in src_aux instead of duplicating the code in ARCore and PAMFD.

ThymoNL (changes requested):

ThymoNL (approved):

PR #778, opened 2022-05-10
Subsystem Failures

abr35:
This is a rough implementation of subsystem failures. Currently it can fail G&N systems in the CM by having circuit breakers randomly fail open at different rates set through the new "SYS" button on the PAMFD failure page. The failures default to "off" and all failures are printed into the Orbiter log. The .Failed valued is placed in the scn second field for each switch to allow for saving and loading.

You can create a lot of contingency situations just by essentially unplugging stuff and it creates the effect of the failure, but it has its limitations (for example Apollo 16's degraded TVC Motor). Doing it another way looks like it would involve a lot more code in each individual subsystem.

Short term to do:
-Allow failed breakers to still be manipulated without supplying voltage (done)
-Place all CSM (done)/LM circuit breakers into a class array to make the code cleaner
-Change the "realistic" failure rate to be realistic based on historical data.

Long term:
-Allow thrusting, extreme maneuvering, or extreme temperatures to effect failure rate (if any data supports this)
-Allow for tanks to leak


Orbitersdk/samples/ProjectApollo/src_sys/toggleswitch.cpp
```diff
@@ -1140,6 +1142,9 @@ bool CircuitBrakerSwitch::CheckMouseClickVC(int event, VECTOR3 &p)
 
 double CircuitBrakerSwitch::Voltage()
 {
+	GetState();
+	if (Failed) return 0.0;			//If a braker has failed, the braker will not supply voltage
+	
 	if ((state != 0) && SRC)
```
indy91: The two lines above are not necessary. Just change "state" to "GetState()" and it does the same thing.

abr35:
GetState() is used properly in Voltage() and the breakers can now be manipulated when failed.

Orbitersdk/samples/ProjectApollo/src_sys/toggleswitch.cpp
```diff
@@ -487,7 +487,7 @@ void TwoPositionSwitch::SaveState(FILEHANDLE scn)
 {
 	char buffer[1000];
 
-	sprintf(buffer, "%i %u", state, GetFlags()); 
+	sprintf(buffer, "%i %i %u", state, Failed, GetFlags()); 		//Saves Failed attribute in second field
```
ThymoNL: You can't change the position of already existing fields. That will break all existing saves.
If you need to introduce a new field, append it instead of inserting. Ideally this would be part of the flags instead of its own thing though.

ThymoNL (changes requested):

Orbitersdk/samples/ProjectApollo/src_sys/toggleswitch.cpp
abr35: I tried adding "fs.Failed = Failed ? 1 : 0;" to GetFlags() and also added it to ToggleSwitchFlags struct, but it only prints 2 fields, what I assume are state and held. Anything I'm missing to get Failed added to the flags?

abr35 (commented):
Moved Failed field to end of string.

n7275:
@abr35 you've pushed an additional commit since marking this ready for review. Do you have more to add or do you want us to look at it now?

ThymoNL:
I've tagged this as a Future Release. While I love that there is the ability to fail circuit breakers I think a more future proof solution that does not depend on a static list of CBs but uses a more generic interface to let systems register their failure states and triggers would be more viable.

This is a good starting point for more advanced systems but I think it needs more time to mature before being merged to mainline NASSP.

indy91:
A bunch of ideas from this PR have been taken up in my own version of a failure simulation overhaul (#1121). As this has become quite dated I will go ahead and close this. Hopefully my version fulfulls most ideas that this PR was meant to improve. Thanks @abr35 !

PR #780, opened 2022-05-18
Fix Apollo 8 MCC bug in the case both MCC-1 and MCC-2 are being scrub…

indy91:
…bed; add a LVDC debug print

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_launch/RTCC_Mission_C_PRIME.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_launch/RTCC_Mission_C_PRIME.cpp
@@ -457,9 +457,11 @@ bool RTCC::CalculationMTP_C_PRIME(int fcn, LPVOID &pad, char * upString, char * 
 		{
 			sprintf(upMessage, "%s has been scrubbed.", manname);
 
-			//Always save table 2 data for MCC-1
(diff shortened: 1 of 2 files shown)


Orbitersdk/samples/ProjectApollo/src_launch/RTCC_Mission_C_PRIME.cpp
```diff
@@ -457,9 +457,11 @@ bool RTCC::CalculationMTP_C_PRIME(int fcn, LPVOID &pad, char * upString, char *
 		{
 			sprintf(upMessage, "%s has been scrubbed.", manname);
 
-			//Always save table 2 data for MCC-1
+			//Always save LOI time and table 2 data for MCC-1, even if the maneuver is scrubbed
 			if (fcn == 20)
```
ThymoNL, line 461: This block is now identical with the one in the else case. So it should be able to move this out of `if (scrubbed)` and simply do `if (fcn == 20 || fcn == 21)`

indy91: I thought about this, but I don't want the LOI time to be stored if MCC-2 is being calculated (fcn == 21), but the maneuver is getting scrubbed.

ThymoNL (approved):

PR #785, opened 2022-05-26
simulate diffusion 

### LINKED ISSUE #783: h_Pipe class does not support diffusion
diffusion rate between two pipe-connected volumes should be proportional to:

someGlobalConstant * (in_t + out_t )/2 *pipe_size * exp(-abs(in_p - out_p))

this would be in an 

if(two_ways)){
    //...
} 

block at the end of void h_Pipe::refresh(double dt)

n7275:
should close #783


n7275:
closing for now. will continue in branch.

PR #787, opened 2022-05-30
RTCC MFD: Fix bug in Angle_Display function by changing variable name  time to angle

indy91:
This caused a Linux only build error, see here: https://www.orbiter-forum.com/threads/linux-playground.40476/post-595561


ThymoNL:
Wow nice bug. Time is not a variable here, that's a comparison to the function pointer of time.
The compiler settings really ought to get reviewed and should fail on warnings to prevent stuff like this.

ThymoNL (approved):

PR #789, opened 2022-06-02
New Docking Initiation Processor for RTCC

indy91:


rcflyinghokie (approved):
Used the processor for Apollo 11 No PDI <60 fps rescue and all the numbers matched up with the rescue plan.

PR #792, opened 2022-06-10
Improve calculation of CSM skin reflection zone for the CSM high-gain antenna

indy91:
There was a rare case of an attitude where the HGA should have still worked, but the skin reflection zone calculation we used indicated that it shouldn't work anymore. The new calculation uses the 45° between CSM X-axis and direction of Earth as the skin reflection zone.

In the future this could be expanded to only prevent acquisition in this zone, as some tracking was still possible in reality.


n7275:
I think this is good for now for the skin reflections. In actuality it would probably cause a bunch of multipath problems relating to the phase of the reflected signal off of the SM skin, and would cause oscillations https://ntrs.nasa.gov/api/citations/19690008842/downloads/19690008842.pdf#page=17 similar to https://youtu.be/EeeeTVr4Lyg?t=776

For some reason the HGA now appears to get stuck up against the scan limit. Maybe it needs a delay before it can acquire again. I will have to look into my notes and files I had saved on this.

n7275 (approved):

PR #793, opened 2022-06-10
Port NASSP to OpenOrbiter

CaptainSwag101:
For the future of NASSP's development, I think it would be wise to start planning to build for OpenOrbiter, since it has several bugfixes and other improvements that may make development easier. Here are what needs to be done to make this happen:

- [x] Fix building with OpenOrbiter's SDK
- [x] Switch from OrbiterSound to XRSound
- [x] Fix remaining XRSound bugs (specific sounds like non-RCS engines were originally handled directly by OrbiterSound, and their functionality will either need to be moved to XRSound or done directly in NASSP now.)
- [ ] Fix certain D3D9Client issues, such as the EMS scroll texture causing crashes in the 3D VC, and the Floodlights not working
- [x] Fix non-looping sounds like switches etc. not working if a user loads into a scenario starting with the 3D VC, right now those sounds don't work until switching to the 2D cockpit at least once per scenario load.
- [x] Fix the 32-bit moons of Mars, Uranus, and Neptune causing corruption and destroying vessel orbits (remove them from our configs)
- [ ] Any other bugs found after the time of writing.


ThymoNL:
I have seen patches by @TheGondos that add CMake support to NASSP. Those will also be a big help to making NASSP's development future proof.

maxq519:
Has this PR been replaced by https://github.com/orbiternassp/NASSP/pull/1288, or is this still being maintained?

CaptainSwag101:
This is no longer being maintained. Thanks for the reminder, I'll close this PR.

Issue #767, opened 2022-04-05
CSM VC Viewpoint bug


maxq519:
Here is the issue:  The CSM VC viewpoint moves when docked with the LM.
I did some testing, and found that it happens in earth orbit, and a little bit in lunar orbit.  Seems like it has something to do with angular accelerations… It only happens when docked with the LM, as soon as I undock the viewpoint jumps back to where it should be.

Absolute Animation Handling is turned on in D3D9 settings.

indy91:
I'm looking a bit into this now. It seems the code that simulates vibrations during launch and maneuver might be responsible. The way it calculates the acceleration acting on the CSM is different from how our accelerometers do it. There seems to be an issue with the force vector returned by Orbiter for docked vehicles. That would explain why it only really happens with CSM+LM docked.

@ggalfi do you want to take a look at this? For testing I let it display the vector vAccel, in Earth orbit with CSM+LM docked.

ggalfi:
Ok, I see now the problem. I think Orbitersim uses some constraint forces to compensate the slightly differing gravitational acceleration of the connected vessels. I think the easiest thing would be to calculate vAccel in the same way as it was done for the PIPAs. My only concern is that doing that doing exactly the same calculation twice (once for the IMU, once for the vibration) is not the nicest thing I could imagine, but anyway it should work.

indy91:
That calculation is in a few more places actually. LV IMU, EMS, G meter in the CSM, T/W meter in the LM, AGS accelerometer.

What might be good is to implement a common class that each vessel can use that calculates the proper acceleration of the vehicle. And then each system that simulates an accelerometer can access that acceleration which only gets calculated once per vessle. But looking at all the accelerometer code we have, it's everywhere slightly different and not everywhere does it use it in the local vessel coordinates. IMU for example directly converts the acceleration from inertial (ecliptic) coordinates to stable member coordinates. So that is quite annoying to find a system that works for all of them...

n7275:
Have we solved this?

ggalfi:
Merging this PR: https://github.com/orbiternassp/NASSP/pull/811 should have solved the issue. I tested it even during Apollo 9's docked DPS burn, and as the acceleration comes from nearly in -X direction (from CSM viewpoint), the COAS remained centered.

PR #795, opened 2022-06-17
Solution for VC head movement bug

ggalfi:
This bug: https://github.com/orbiternassp/NASSP/issues/767 was caused because GetForceVector does not take into account all the forces when the vessel is docked. A new class InertialData is created to correctly calculate inertial acceleration. The head movement in VC is now based on this instead of GetForceVector function.
This - or a very similar - calculation is done many times for multiple instruments (IMU, EMS, mechanical G-meter on CSM, LV IMU, LM IMU, LM T/W, LM AGS Accelerometer). It would be more effective to make all these to query the InertialData object of the vessel for inertial acceleration instead of doing the calculation on their own.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_sys/inertial.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_sys/inertial.cpp
@@ -1,0 +1,148 @@
+/***************************************************************************
+This file is part of Project Apollo - NASSP
+Copyright 2004-2022
+
+Generic Apollo Guidance computer class which is hooked
+up to the DSKY to allow easy support for the CSM and LEM
+computers. This defines the interfaces that the DSKY will
+use to either computer, so only parts specific to the CSM
+or LEM will need to be written specially.
+
+Project Apollo is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 2 of the License, or
+(at your option) any later version.
+
+Project Apollo is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with Project Apollo; if not, write to the Free Software
+Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+See http://nassp.sourceforge.net/license/ for more details.
+
+**************************************************************************/
+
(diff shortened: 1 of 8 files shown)


indy91:
I'll give this a good test. Looks good though! Something like this was also my idea, but when I started looking into it I was a bit discouraged by how different the accelerometer implementations of LV IMU, EMS, G-Meter etc. are. I had hoped to implement something that generated the data that all accelerometers could access. But that can hopefully still be done with this, just will need some changes in those accelerometer classes.

ggalfi:
> I had hoped to implement something that generated the data that all accelerometers could access. But that can hopefully still be done with this, just will need some changes in those accelerometer classes.

My observation was when I peeked into the code of these instruments was that a lot of unnecessary calculation is happening (on top of the repeated calculation of acceleration): for example in case of EMS to get the the vessel->global transformation and it's inverse, there is a lot of messing with Euler angles and also trigonometric functions - but we don't need them at all. First of all, the actual attitude of a vessel is kept as a rotational matrix within Orbiter (I've looked it up in Orbiter's source code), so I'd rather query it directly by `GetRotationMatrix `method than to ask OrbiterSDK to convert it into angles through `GetGlobalOrientation`. Also to get the inverse of this transformation, there is no need to mess with angles, as it is a rotational matrix it's inverse is simply it's transpose. And despite it is fairly easy, still you don't need to calculate the transpose itself, just use the `tmul` function instead of `mul` to multiply with the inverse of the matrix.
If you don't mind I'd keenly do the integration of all the relevant instruments with IntertialData class, and I'd also do some refactoring of the linear algebra stuff there.

indy91:
Yeah this new inertial data class seems like a good place to do a bit of experimentation like that. Go for it!

One thing to keep in mind. When we originally switched to Orbiter 2016 we suddenly had a small oscillation in the acceleration. I believe this was caused by using the wrong Orbiter timestep length in clbkPostStep because the way clbkPostStep works changed in Orbiter 2016. It probably doesn't occur in your class (I haven't checked yet), but just something to be careful about. Our EMS didn't like it and triggered 0.05g too early for example.

Orbitersdk/samples/ProjectApollo/src_sys/inertial.cpp
```diff
@@ -0,0 +1,148 @@
+/***************************************************************************
+This file is part of Project Apollo - NASSP
+Copyright 2004-2022
+
+Generic Apollo Guidance computer class which is hooked
```
ThymoNL, line 5: Don't forget to update this description. Also since this is a new file Copyright would be just for this year and not back to 2004. ;)

PR #796, opened 2022-06-22
RTCC bug fixes

indy91:
-SPS thrust was hardcoded in P30 REFSMMAT calculation. Now can also use DPS
-More error messages for Return to Earth Digitals
-P30 uplink with MPT active can uplink CSM MPT maneuver to LGC and vice versa
-General purpose maneuver to change apo/peri adjusts apo or peri if radius before maneuver is out of range
-Vector conversion routine outputs reference body indicator


ThymoNL (approved):
lgtm

PR #797, opened 2022-06-23
Updated gravity models

indy91:
-Adds third harmonic of lunar potential function (J3) to Moon config file and RTCC functions. It was previously missing because it was based on the Apollo 8, Colossus 237 GSOP. Using an Apollo 11 state vector the addition of J3 only makes a difference of 70 meters in 24 hours
-Adds (commented) Earth and Lunar potential coefficients that Orbiter cannot simulate yet. Brings the RTCC to the state in the late Apollo program
-To make use of these coefficients, the Shuttle onboard navigation equations have been partially implemented, which should be an efficient implementation 
-Delete unused, experimental coasting integration function
-All the Orbiter rotational functions have been streamlined and made more efficient, as they are now used more in the RTCC gravity calculations


n7275 (approved):
Tested against Apollo 10 scenerios without issue.

PR #799, opened 2022-06-30
Run systems simulation on the first timestep in CSM and LM

indy91:
Previously very little code was run in the CSM and LM on the first timestep after loading into an Orbiter session. This first timestep always has a duration of 0.01 seconds. This time was not accounted for in the simulation of any time keeping code, which includes the MissionTimer variable (used to show mission time in the PAMFD among many other things) but also the AGC. No AGC cycles are done during this initial 0.01 seconds. That means every time you are loading an Orbiter session while flying a mission you give the AGC an additional clock error of 0.01 seconds. Do this very often during a mission and you could get a significant clock drift.

The reason for not calling the main timestep functions in CSM and LM was given in a code comment:

> On the first timestep we just do basic setup stuff and return. We seem to get called in at least some cases before Orbiter is properly set up, so the last thing we want to do is point the engines in a wacky direction and then not be called again for several seconds.

It is quite possible that this issue is long gone and doesn't apply to the current Orbiter anymore. It could very well be that it was already fixed in Orbiter 2010! This PR has no problem loading scenarios, but I think it's still a significant change that could lead to new bugs. So it can use a bunch of testing!

LastTimestep was just an unused variable related to the first timestep code.


rcflyinghokie (approved):
I didn't notice any ill effects but I also play on a more powerful machine.

ThymoNL (approved):

PR #811, opened 2022-08-02
New inertial acceleration calculation logic

ggalfi:
Added new class to handle inertial acceleration. AGC's IMU, LV IMU, EMS, AGS, MechanicalAccelerometer and POV movement visuals are now fed from this new class. Tested with a lot of different scenarios where acceleration measurements are important: launch, entry, SPS burns, P47, P57, lunar ascent and rendezvous. I haven't seen any problem with the new logic of acceleration calculation.


indy91 (changes requested):
Great work! I've tested a bunch of things and it all worked very good. Docked burn, no oscillation in the acceleration (we had this when we switched to Orbiter 2016, can happen if you use the current simdt instead of lastSimDT).

I found only one relevant issue. The S-IVB class also needs to generate the intertial data. And then it needs the same code for the IULV_GET_INERTIAL_ACCEL message that you also added in csmconnector.cpp.

I think there is only one instance of the separate S-IVB needing these acceleration data, which is the Apollo 9 "TLI" burn that the S-IVB does on its own. That would be broken with your PR as-is. But it should be an easy fix.

ggalfi:
> I found only one relevant issue. The S-IVB class also needs to generate the intertial data. And then it needs the same code for the IULV_GET_INERTIAL_ACCEL message that you also added in csmconnector.cpp.

I've added InertialData and all the necessary stuff to the SIVB class. Tested on Apollo 9 with the first standalone S-IVB burn occurring at GET 4:45:41, and it put the vehicle on a H_apo ~ 3100km orbit.

indy91:
Works great now! There is just one problem now, as you can see below there is a merge conflict in the Visual Studio files for the Saturn1B and Saturn5NASP modules. Your files for the inertial data class were added and something must have been changed since you branched off from the Orbiter2016 branch. I think all you need to do is merge the current Orbiter2016 branch back into your accel branch and then you need to solve the merge conflict with some annoying editing of those VS files.

If you want I could try to do that as well. Not sure how to directly amend your PR, but I could probably create a PR for your branch. So layers of PRs :D

indy91:
I think I can fix it for you, standby.

indy91 (dismissed):

indy91:
Seems to work now. @ThymoNL can you take a look that I didn't screw something up, I didn't really know what I was doing and committed on a detached head.

ThymoNL:
> Seems to work now. @ThymoNL can you take a look that I didn't screw something up, I didn't really know what I was doing and committed on a detached head.

Sure, I still need to do an in depth review either way. I'll be back from vacation on the 11th. Unless someone is urgently waiting on this I'll look at it next week.

Orbitersdk/samples/ProjectApollo/src_csm/scs.cpp
```diff
@@ -5140,7 +5137,7 @@ void EMS::SystemTimestep(double simdt) {
 }
 
 void EMS::AccelerometerTimeStep(double simdt) {
-
+	/*
```
ThymoNL: If you comment out code you can just as well remove it. If this change needs to be undone in the future we can revert the commit. I see no reason to keep this.

Orbitersdk/samples/ProjectApollo/src_sys/inertial.h
```diff
@@ -0,0 +1,48 @@
...
+	VESSEL *vessel;
+	bool dVInitialized;
+	VECTOR3 lastWeight;
+	VECTOR3 lastGlobalVel;
+	double lastSimDT;
+
+public:
+	InertialData(VESSEL *vessel);
+	VECTOR3 GetGravityVector();
+	void timestep(double simdt);
```
ThymoNL: Pardon me for being pedantic. To be consistent with other objects handling timestep the T would be capitalized.

Orbitersdk/samples/ProjectApollo/src_sys/inertial.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+void InertialData::getAcceleration(VECTOR3 &acc) {
+	acc.x = accel.x;
+	acc.y = accel.y;
+	acc.z = accel.z;
+}
+
+void InertialData::SaveState(FILEHANDLE scn){
+	oapiWriteLine(scn, INERTIAL_DATA_START_STRING);
+	oapiWriteScenario_int(scn, "DVINITIALIZED", (dVInitialized ? 1 : 0));
```
ThymoNL, line 150: I know this was already saved in the previous implementation. However, would not saving these and instead reinitializing them on each load cause any problems? It would remove four variables from the ever increasing list of things to save.

ThymoNL (changes requested):

ThymoNL:
See comments. Changes look good otherwise! Thanks

Orbitersdk/samples/ProjectApollo/src_sys/inertial.cpp
indy91: Maybe there is a way, but I think the problem is that, despite the timestep function being called in clbkPostStep, in Orbiter 2016 the simulation has changed so that functions like GetWeightVector aren't updated until AFTER clbkPostStep, that was one of the first issues to fix in NASSP when we moved to 2016. See this thread: https://www.orbiter-forum.com/threads/vessel-getglobalvel-and-or-simdt-oscillation.34756/#post-533107

So when this timestep function is being called for the first time after starting the sim, the weight vector etc. are still identical to the values that Orbiter would return during your proposed reinitialization function call, which would be during scenario loading. So I don't really see a way to do it unless we change how our accelerometers work in general.

Also, this PR will actually reduce the number of scenario lines, because each accelerometer class (IMU etc.) won't need to save this data anymore. Just look how many oapiWriteScenario/papiWriteScenario are being deleted!

ThymoNL: Sounds reasonable. Just making sure it can't be optimized away all together. Let's keep it then.

ggalfi:
Hi all, I have reformatted the code according to Thymo's request. Per your discussion I haven't touched the saving-loading functionality of InertialData object.

indy91 (dismissed):
Looks perfect to me! Please have a look @ThymoNL and if you like it, then it's finally time to merge this.

Orbitersdk/samples/ProjectApollo/src_csm/scs.h
```diff
@@ -859,7 +859,7 @@ class EMS : public e_object {
 	bool IsDisplayPowered();
 	
 	void AccelerometerTimeStep(double simdt);
-	VECTOR3 GetGravityVector();
+	//VECTOR3 GetGravityVector();
```
ThymoNL: commented

Orbitersdk/samples/ProjectApollo/src_csm/scs.h
```diff
@@ -868,10 +868,10 @@ class EMS : public e_object {
 	int GScribe; //pixels
 	double ScrollPosition; //fractional pixels
 	double MaxScrollPosition;
-	bool dVInitialized;
+	/*bool dVInitialized;
```
ThymoNL: commented through ln874

ThymoNL (commented):
There's two commented blocks remaining. Looks great otherwise!
@indy91 Feel free to merge once those remaining blocks have been taken care of.

indy91:
I'll take care of fixing the conflict, like last time. I can also remove the commented lines. Stay tuned...

indy91 (approved):

ThymoNL (approved):

PR #819, opened 2022-08-11
S-band Ground station antenna tracking.

n7275:


Orbitersdk/samples/ProjectApollo/src_launch/mccvessel.h
```diff
@@ -24,6 +24,7 @@
 #pragma once
 
 #include "Orbitersdk.h"
+#include "connector.h"
```
indy91: Is this needed or is it a remnant before you started using clbkGeneric?

Orbitersdk/samples/ProjectApollo/src_csm/csm_telecom.cpp
```diff
@@ -1010,23 +1011,29 @@ void HGA::TimeStep(double simt, double simdt)
...
+	//Global position of the spacecraft, spacecraft rotation matrix from local to global
 	sat->GetGlobalPos(pos);
-	oapiGetGlobalPos(hEarth, &R_E);
-	oapiGetGlobalPos(hMoon, &R_M);
 	sat->GetRotationMatrix(Rot);
+
+	//Get the gain, power and global position of the transmitter
+	MCCVessel->clbkGeneric(paCBGmessageID::messageID::RF_PROPERTIES, paCBGmessageID::parameterID::GetTxPosition, &R_E);
+	MCCVessel->clbkGeneric(paCBGmessageID::messageID::RF_PROPERTIES, paCBGmessageID::parameterID::GetTxPower, &TransmitterPower);
+	MCCVessel->clbkGeneric(paCBGmessageID::messageID::RF_PROPERTIES, paCBGmessageID::parameterID::GetTxGain, &TransmitterGain);
```
indy91: Maybe this could be made more efficient by putting all three of these values in a struct that the void point is referencing, so sending this data as a package instead of three separate clbkGeneric calls. Just an idea though.

Orbitersdk/samples/ProjectApollo/src_launch/mccvessel.h
n7275: That's a remnant. I will remove it.

Orbitersdk/samples/ProjectApollo/src_csm/csm_telecom.cpp
n7275: I think I had something similar when I was using connectors. Rf_properties or something like that. Good idea.

indy91 (approved):

PR #809, opened 2022-07-26
Implement LM master alarm buttons as actual PanelSwitchItem, usable by the Checklist MFD

indy91:

PR #823, opened 2022-08-15
Add CWEA reset MA pushbutton pushes to LM checklists

rcflyinghokie:
This will work in conjunction with #809


rcflyinghokie:
Changes in incorrect branch, will push in checklist branch

PR #828, opened 2022-08-21
TLI on MPT bugs

indy91:
-During trajectory update, the new TLI ignition time is stored correctly
-Initialize TLI ullage DV counter to zero (could be NaN otherwise)
-Implement a small tolerance on double comparison in interpolation routine


n7275 (approved):
this fixes my issue with the Next Station Contacts display.

PR #834, opened 2022-08-28
Fuel cell nitrogen pressurization system and telemetry words

n7275:
needs some long-term testing to make sure that the header tanks don't run out on a full mission. Currently there is no consequence of this happening, but it should be thoroughly tested before the fuel cells depend on it.


n7275:
This is now part of #906 and will get merged there, closing.

PR #840, opened 2022-09-05
Refined Q-ball calculation

ggalfi:
Simulation of Q-ball now takes into account the dynamical pressure and AOA together, which is more closer to reality than the original, only AOA-based calculation (Q-ball is practically a pressure measuring device). Currently we assume that the delta pressure returned by the Q-ball is proportional to the product of AOA and dynamical pressure. The coefficient of this relation could be estimated based on the  Launch Vehicle Flight Evaluation Reports, see comments for details. As the max delta pressure happened mostly in the supersonic region, one may try to use the more precise Rayleigh pitot tube formula (see e.g. here: https://ntrs.nasa.gov/api/citations/20200002412/downloads/20200002412.pdf), but that gave me less consistent results with the Evaluation Reports, so I decided to stick with the simple aoa*q expression.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_saturn/qball.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_saturn/qball.cpp
@@ -51,9 +51,41 @@ double QBall::GetAOA()
 		{
 			aoa = 10.0*RAD;
 		}
-		else if (sat->GetDynPressure() > 100.0)
-		{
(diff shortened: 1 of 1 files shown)


n7275 (approved):
I flew two launches with this change. Looks good.

indy91:
Yes, this makes a lot of sense. Total angle of attack.

PR #852, opened 2022-10-06
Fix launch FC ties for CSM Default Checklist

rcflyinghokie:


rcflyinghokie:
Launch configuration is correct

PR #854, opened 2022-10-08
RTCC MFD: LS REFSMMAT requires CSM state vector; AGS SV PAD has speci…

indy91:
…al logic for landed LM


rcflyinghokie (approved):
From testing on the surface, this rounds properly and pulls the CSM REFSMMAT allowing the user to generate a LM AGS SV pad without extra steps of pulling REFSMMAT between vehicles or playing roulette with the CLC button.

PR #860, opened 2022-10-25
Change LGC version used for Apollo 13 from Luminary 131 to LM131R1

indy91:


indy91:
List of padload changes:

MASS (1243 and 1244). Now identical to actual padload. Double precision mass, but only most significant word is loaded in the actual padload. Values are updated during LM activation, so precise value loaded here doesn't matter much.

LEMMASS (1326): Now identical to actual padload. Mass that appears in V48. See above.

CSMMASS (1327): Now identical to actual padload. Mass that appears in V48. See above.

IMU biases (1452 to 1470). Previously not loaded at all, now as all zeros, which means the same thing.

TEPHEM+2 (1710): Now identical to actual padload. Previous value had a typo (33650 instead of 33560), but value gets updated during LM activation.

AZO+1 (1712) and AXO+1 (1716). One bit changes to Earth polar axis correction values, slightly different because of a different time for which they are calculated. Inconsequential difference.

504LM (2012 to 2017): Moon libration vector, calculated to give the best accuracy for the lunar landing site for inertial to planetary coordinate transformations. Previously calculated using the wrong LGC constants that were used in Luminary 99, but got correct in Luminary 116.

AAPFG+1 (2463): Now identical to actual padload. The last digit was difficult to read in padload document, should be a 6, but was read as a 5.

TAUHZ (2510), QHZ (2511), AHZLIM (2512), TOOFEW (2513), HLROFF (2514 and 2515), 2LATE466 (2516 and 2517). These are the changes for the Auto P66 and the only changes strictly required to run Luminary 131R1 instead of 131. Now identical to actual padload.

ABTRDOT (2567): Now identical to actual padload. Another 6 that was mistaken for a 5. Again an inconsequential one bit difference.

DELTTFAP+1 (3425): Now identical to actual padload. Accidentally transposed digits. Value is the change in time-to-go when the descent guidance switches from P63 to P64. Old value did not seem to negatively impact the behavior too much, only the least significiant two digits in a double precision word were slightly wrong here.

thewonderidiot (approved):
All looks good to me!

PR #869, opened 2022-11-04
LVDC: Venting and drag calculations fully implemented (but not fully …

indy91:
…used yet); significantly reduce number of debug lines written to the log


rcflyinghokie (approved):
Didnt see any issues with this, nothing breaking when utilized and less debug lines is probably good :)

PR #870, opened 2022-11-05
Fix jump in camera position in the LM virtual cockpit at gear deploym…

indy91:
…ent and staging


rcflyinghokie (approved):
Tested on Apollo 10, didn't see a jump in VC

PR #890, opened 2023-01-06
Migration to sketchpad API

TheGondos:
This is work to switch graphics from the GDI API to the sketchpad API.
There are some visual differences as noted here with the D3D9Client (these are animated gif, you may have to zoom in a little to see what happens...)
- ~~On the EMS the "RSITriangle" is fatter on GDI and there's some kind of shadowing effect~~ (fixed) :
EMS
- The text on the MFD buttons is also shifted for some texts :
MFD-v2

Any ideas about the cause of these differences? Are they significant enough to be a problem?

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/satswitches.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/satswitches.cpp
@@ -588,7 +588,7 @@ void SaturnPartPressCO2Meter::DoDrawSwitch(double v, SURFHANDLE drawSurface)
 		oapiBlt(drawSurface, NeedleSurface, 215, (20 - (int)((v - 20.0) / 10.0 * 14.0)), 10, 0, 10, 10, SURF_PREDEF_CK);
 }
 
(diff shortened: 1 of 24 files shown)


TheGondos:
hum, there is a crash on the EMS scroll when switching to the 3D cockpit. It maybe be a bug in the orbiter core : I switched to a 'dynamic' load (`srf[SRF_VC_EMS_SCROLL_LEO] = oapiLoadTexture("ProjectApollo/VC/EMSscroll_LEO.dds", true);`) but it still fails.
When looking in OrbiterAPI.cpp there is this :
```
DLLEXPORT SURFHANDLE oapiLoadTexture (const char *fname, bool dynamic) {
	oapi::GraphicsClient *gc = g_pOrbiter->GetGraphicsClient();
	if (gc) return gc->clbkLoadTexture (fname, 8 & (dynamic ? 3:0));
	else return NULL;
}
```
```8 & (dynamic ? 3:0)``` will always resolve to 0 and should probably be ```8 | (dynamic ? 3:0)```
Edit: worked around

indy91 (approved):

PR #891, opened 2023-01-07
Provide safe Watt to dBm conversion

TheGondos:
With FPU exceptions on, the conversion to dBm can stop the program because taking the log10 of 0 triggers an exception.
This PR proposes a new function to do the conversion safely by checking the power beforehand and returning RF_ZERO_POWER_DBM if it is zero.
Edit: forced pushed an amend to use RFCALC_W2dBm in RFCALC_rcvdPower.


n7275 (approved):
I had good results on everything I tested. This should be okay to merge.

PR #892, opened 2023-01-08
Use C++ thread API

TheGondos:
This replaces the windows API with standard C++ threads.
I'd need some pointers on how to properly test this because I'm clueless about the MCC stuff^^

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_launch/mcc.h
+++ b/Orbitersdk/samples/ProjectApollo/src_launch/mcc.h
@@ -295,7 +295,7 @@ public:
 	// SUBTHREAD MANAGEMENT
 	int subThreadMode;										// What should the subthread do?
 	int subThreadType;										// What type of subthread?
-	int subThreadStatus;									// 0 = done/not busy, 1 = busy, negative = done with error
+	std::atomic<int> subThreadStatus;						// 0 = done/not busy, 1 = busy, negative = done with error
 
 	// GROUND TRACKING NETWORK
 	struct GroundStation GroundStations[MAX_GROUND_STATION]; // Ground Station Array
--- a/Orbitersdk/samples/ProjectApollo/src_launch/mcc.h
+++ b/Orbitersdk/samples/ProjectApollo/src_launch/mcc.h
@@ -27,6 +27,7 @@
 #include "MCCPADForms.h"
 #include <fstream>
 #include <atomic>
+#include "thread.h"
 
 // Save file strings
 #define MCC_START_STRING	"MCC_BEGIN"
@@ -296,7 +297,7 @@ public:
 	// SUBTHREAD MANAGEMENT
 	int subThreadMode;										// What should the subthread do?
 	int subThreadType;										// What type of subthread?
-	std::atomic<int> subThreadStatus;						// 0 = done/not busy, 1 = busy, negative = done with error
+	std::atomic<ThreadStatus> subThreadStatus;
 
 	// GROUND TRACKING NETWORK
 	struct GroundStation GroundStations[MAX_GROUND_STATION]; // Ground Station Array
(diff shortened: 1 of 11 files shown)


Orbitersdk/samples/ProjectApollo/src_launch/mcc.h
```diff
@@ -295,7 +296,7 @@ class MCC {
 	// SUBTHREAD MANAGEMENT
 	int subThreadMode;										// What should the subthread do?
 	int subThreadType;										// What type of subthread?
-	int subThreadStatus;									// 0 = done/not busy, 1 = busy, negative = done with error
+	std::atomic<int> subThreadStatus;						// 0 = done/not busy, 1 = busy, negative = done with error
```
ThymoNL: Don't use integers to indicate a status but add them to an enum instead.

ThymoNL (commented):
Looks sane at first glance. I'll give this a test this weekend.

Orbitersdk/samples/ProjectApollo/src_launch/mcc.h
TheGondos: I can do that but I'd recommend doing the same in ARCore to keep things aligned.
Also I have not seen a single instance of subThreadStatus being set to a negative value so the comment is likely wrong ; however there is a '2' value to indicate it is running ('1' is kind of a 'scheduled but not yet running' state).
Do we want to have the option to specify different error codes or just a single value indicating an unspecified error? (it would be easier to use an enum class this way).
I would go with this :
```
enum class ThreadStatus {
   Error,
   Stopped,
   Scheduled,
   Running
};
```
What do you think?

dseagrav:
I wrote the comment with the expectation that a negative number would indicate error so a success test could be simply “x > 0” and a failure test simply “x < 0” without having to test for every possible error state. It was for future use.

Please keep errors testable by simple means. Consider that we run this code inside of orbiter’s timestep, and any non-trivial operations in a timestep cost framerate, which has other negative impacts.

Everything in a timestep should be kept as computationally inexpensive as possible, even if that means we do not adhere to “best practices” by C++ pattern conventions. The project is worthless if it the code runs too slowly for use.

Orbitersdk/samples/ProjectApollo/src_launch/mcc.h
TheGondos: Pushed a modification for this. An enum is used instead of an int, the change has been done in the ARCore too to keep things the same.

Orbitersdk/samples/ProjectApollo/src_sys/thread.cpp
```diff
@@ -23,22 +23,37 @@
...
-DWORD WINAPI Runnable::ThreadEntry (void* arg)
-{
-    Runnable * pRunnable = (Runnable *) arg;
-    pRunnable->Run();
-    return 0;
+// Some RTCC algorithms can loop forever if there is a convergence problem,
+// so we need a way to kill the worker thread.
+// Unfortunately there's no standard way to do that so we have to revert to
+// native thread API to do it
+void KillableWorker::Kill() {
```
ThymoNL, line 47: This method leaks resources. I would add a comment that notifies the user of that fact so that this doesn't get used without explicitly knowing you are causing a memory leak. This should only be used if there is no other way to safely terminate a thread.

I do agree that there is no platform independent approach that covers this nicely.

ThymoNL (approved):
Tested this by flying a launch and letting MCC do some calculations. Seems to work great!
I have left one note regarding Kill.

Thanks so much for making NASSP a little bit more platform independant.

ThymoNL (changes requested):
see above comment

Orbitersdk/samples/ProjectApollo/src_sys/thread.cpp
TheGondos: Comment added, hope it's not too dramatic :grin:

ThymoNL: I think that gets the message across. Thanks!

ThymoNL (approved):

Issue #861, opened 2022-10-25
[VC][CSM]Calibrate H2 and O2 pressure indicators


TurryBoeing:
On the VC, the pressure gauges for H2 and O2 don´t match the 2D needle positions/values.
![cryo press issue 1](https://user-images.githubusercontent.com/7021346/197855215-5f6410f1-6529-453f-ac43-27da4beb000a.png)
![cryo press issue 2](https://user-images.githubusercontent.com/7021346/197855237-c8576c15-912d-4b06-9766-42c3c4c3f757.png)

n7275:
@indy91, this can be closed right?

indy91:
@n7275 Yes, @TurryBoeing confirms it is fixed now.

PR #897, opened 2023-01-23
VC: Fix alignment of H2 and O2 pressure meters in CSM and steerable a…

indy91:
…ntenna yaw meter in the LM; measure cryo pressures with sensors


indy91:
For issue #861

n7275 (approved):

PR #899, opened 2023-01-23
Rendezvous radar stays inertially fixed in slew mode

indy91:


rcflyinghokie:
Other than it taking some getting used to, this has worked well in practice, I would like to complete my Apollo 12 rendezvous with it but I think it is pretty safe to say its very mergeable

rcflyinghokie (approved):
I have flown this on Apollo 12 rendezvous a few times as well as other rendezvous with no ill effects.

PR #906, opened 2023-01-31
Liquid density

n7275:
This model improves the calculation of liquid densities for:

- [X] Hydrogen
- [X] Oxygen
- [x] Nitrogen
- [ ] Helium

This is necessary for more accurate CSM cryogenic tank temperatures over the course of a full mission, additionally this will also allow the existence (and realistic behavior) of SIVb and SII fuel and oxidizer tanks, down the road.

This branch contains commits from two other branches. a rebase when those get merged will fix things.

This will also break older scenarios. I will need to come up with a solution....and be confident that I won't need to do this again for a long time as a result of these changes.


n7275:
my branch https://github.com/n7275/NASSP/tree/liquidDensity2 will be the one be to PRed this will now be closed, but the branch will remain as a diff check.

PR #907, opened 2023-02-07
Fc temp meter

### LINKED ISSUE #900: Calibrate VC Fuel Cell Temp Meter Ranges
Fuel Cell temperature meters are not scaled properly to the variable scale ranges on the gauge.

https://github.com/orbiternassp/NASSP/blob/f3a37083520889feba40739a324e5baa0ef28432/Orbitersdk/samples/ProjectApollo/src_csm/satswitches.cpp#L469

n7275:
Closes #900

Also adds a few extra transducers and wires them to the proper power source as specified in the CSM 114 and 104 systems handbooks.


indy91 (changes requested):
I believe the fuel cell 2 and 3 sensors should be wired to CB3 and not CB4

indy91 (approved):

PR #916, opened 2023-02-22
Fix invalid assignment in if statement

TheGondos:
This fixes a typo in LOITargeting.cpp


indy91:
I already fixed this in #915 so I am a bit confused why this even shows up as a change under "Files changed"...

TheGondos:
> I already fixed this in #915 so I am a bit confused why this even shows up as a change under "Files changed"...

lol, I'll close it then, maybe I forgot to rebase so the diff is from before your modification...

PR #918, opened 2023-02-22
Use InitModule/ExitModule for LEM and saturn

TheGondos:
The LEM and saturn modules are using DllMain to handle sketchpad ressources.
This commit changes that to use the standard InitModule/ExitModule API.
I set it as a draft for now in case someone can remember if there is a valid reason for using DllMain

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
@@ -21,6 +21,7 @@
   See http://nassp.sourceforge.net/license/ for more details.
 
   **************************************************************************/
+#define ORBITER_MODULE
 
 // To force Orbitersdk.h to use <fstream> in any compiler version
 #pragma include_alias( <fstream.h>, <fstream> )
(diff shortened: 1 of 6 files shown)


indy91:
Just finding the time to look at this PR. Shouldn't it be

`DLLCLBK void ExitModule(HINSTANCE hModule)`

instead of

`DLLCLBK void ExitModule()`

I haven't checked yet if it would call ExitModule(), but in other modules I see it with the HINSTANCE.

TheGondos:
> Just finding the time to look at this PR. Shouldn't it be
> 
> `DLLCLBK void ExitModule(HINSTANCE hModule)`
> 
> instead of
> 
> `DLLCLBK void ExitModule()`
> 
> I haven't checked yet if it would call ExitModule(), but in other modules I see it with the HINSTANCE.

That is correct, I'll fix it

indy91:
Thanks for the fix! I had actually checked in the meantime, it was still calling ExitModule despite the difference. But this is better.

indy91 (approved):

Issue #935, opened 2023-03-27
LM RCS issues with Vesim enabled


n7275:
This issue first appeared at #918, the previous merge, #924 (9cce195445f1bc9a668ef269de1ce64b916aa765) does not have the symptoms described below:

- Continuous firing of LM RCS when Vesim is enabled. This happens irrespective or whether or not a joystick is plugged in. (vesim.getInputValue() is always returning 0)
- No issues with CM/CSM RCS, and Vesim works as intended.
- The above two points are still true in scenarios with a LM and CSM, or just a LM or just a CSM.
- When closing a scenario https://github.com/orbiternassp/NASSP/blob/3d997a61397ebe47ac0c8e39f6d39ebf95978f28/Orbitersdk/samples/ProjectApollo/src_lm/LEM.cpp#L507 causes Orbiter to crash during shutdown because of an access violation.
- @indy91 has confirmed the issue as well

@TheGondos, could you help look into this if you have a few minutes? I'm not having any luck with it and I can't see what difference between LM and saturn class could be causing one to work and the other to not work.

PR #938, opened 2023-03-29
Only keep one g_Param in LEM class; fixes issue #935

indy91:


n7275 (approved):
This works! Good to merge.

PR #952, opened 2023-04-17
Improve Model of Liquid Density For Cryogenic Substances

n7275:
Before merging:

- [ ] Create Update script and rev scenario version number.
- [ ] Update Mission scenarios
- [ ] Infor users on through public channels

New Model:
image

https://gist.github.com/n7275/8676c064fef5f48680b5ba815f24e2bf


n7275:
This is now included in #1023

PR #954, opened 2023-04-20
Cue card fixes

indy91:
-Change method of loading parameters from mission files to prevent workarounds being necessary for loading multiple cue cards
-Fix bug with cue card loading from mission file
-Add the code necessary in the future to support LM cue cards
-Show Apollo 14 and 15 boost cards for other missions
-Delete pre-entry attitude timeline card because it couldn't be attached to velcro

Especially the first point is a big change. It was necessary because previously the family of Orbiter functions "oapiReadItem" were used to read lines from the mission files. These functions have the limitation of searching for one specific line in the file, which means you can't have the same item twice. This was a big annoyance with the cue cards, because bad code was necessary to load multiple cue cards from the file ("CSMCueCard1=", "CSMCueCard2=" instead of multiple "CSMCueCard=" had to be used). Now there is no limit on the number of cue cards you could put in mission files.

This branch could use extra testing and checking of the code because it changes the way every single mission specific setting is loaded. I changed it in a backwards compatible way, hence the "TRUE" checks instead of 0/1 in the file. Also @TheGondos might dislike this because it adds more Windows only code (_strnicmp), but what can one do...


n7275 (approved):
From what I was able to test, this appears to be good.

PR #970, opened 2023-05-08
RJEC bias voltage not required for proper operation of solenoid drivers

indy91:


rcflyinghokie:
This now allows hot fire in accel command using the auto rcs coils. Looks like this fixes the behavior issue.

rcflyinghokie (approved):
This now allows hot fire in accel command using the auto rcs coils. Looks like this fixes the behavior issue.

PR #971, opened 2023-05-09
MPT mass update function doesn't try to look for SM propellants after…

indy91:
… the SM was jettisoned; splashdown update page defaults to mission specific reentry range


rcflyinghokie (approved):
Fixes CTD in the case of no SM when computing maneuver.

PR #972, opened 2023-05-10
Fix DSKY power distribution in CSM & LM

rcflyinghokie:
This changes power flow to the DSKY in the CSM and LM to use the appropriate sources. Previously, the CSM incorrectly used the CM AC bus for DSKY power where power for the segment was generated internally.

This adds power mergers (for now) to allow the GN MNA&B breakers in the CM power the DSKY segmented lighting and the AC bus power the status light. In the LM, the same AC internal generation is done via the LGC/DSKY cb for the segmented displays, but status lights now correctly use the DC LTG NUM DOCK COMP cbs on a power merger until the LCA is better implemented.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/satsystems.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/satsystems.cpp
@@ -249,6 +249,8 @@ void Saturn::SystemsInit() {
 	FlightBus.WireTo(&FlightBusFeeder);
 	Panelsdk.AddElectrical(&FlightBus, false);
 
+	CMCDCBusFeeder.WireToBuses(&GNComputerMnACircuitBraker, &GNComputerMnBCircuitBraker);
+
 
 	// Feeder for LM umbilical
 	LMUmbilicalFeeder.WireToBuses(&MnbLMPWR1CircuitBraker,&MnbLMPWR2CircuitBraker);
--- a/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
@@ -362,7 +362,8 @@ Saturn::Saturn(OBJHANDLE hObj, int fmodel) : ProjectApolloConnectorVessel (hObj,
 	inertialData(this),
 	agc(soundlib, dsky, dsky2, imu, scdu, tcdu, Panelsdk),
 	dsky(soundlib, agc, 015),
-	dsky2(soundlib, agc, 016), 
+	dsky2(soundlib, agc, 016),
+	CMCDCBusFeeder("CMC-DCBus-Feeder", Panelsdk),
 	imu(agc, Panelsdk, inertialData),
 	scdu(agc, RegOPTX, 0140, 2),
 	tcdu(agc, RegOPTY, 0141, 2),
@@ -782,8 +783,8 @@ void Saturn::initSaturn()
 
 	agc.ControlVessel(this);
 	imu.SetVessel(this, false);
-	dsky.Init(&LightingNumIntLMDCCB, &NumericRotarySwitch);
-	dsky2.Init(&LightingNumIntLEBCB, &Panel100NumericRotarySwitch);
+	dsky.Init(&LightingNumIntLMDCCB, &CMCDCBusFeeder, &NumericRotarySwitch);
+	dsky2.Init(&LightingNumIntLEBCB, &CMCDCBusFeeder, &Panel100NumericRotarySwitch);
 
 	//
 	// Configure SECS.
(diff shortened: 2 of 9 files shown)


indy91:
Looks good to me. The only small tweak I would like to see is the StatusPower and SegmentPower pointers being nulled in the DSKY constructor. Or else if something wrong is done with the Init function, like it not being called, random behavior would ensue.

rcflyinghokie:
> Looks good to me. The only small tweak I would like to see is the StatusPower and SegmentPower pointers being nulled in the DSKY constructor. Or else if something wrong is done with the Init function, like it not being called, random behavior would ensue.

Completed

indy91 (dismissed):

ThymoNL:
We should only be passing breakers to systems and do any power merges there if they get fed from both main buses. Creating new powermerges outside that system will add unnecessary clutter to the LEM and Saturn code that is hard to maintain.

Can you move the Powermerge to the DSKY class and wire it in the Init function? 

I also want to reinforce that it is good practice to keep your commits atomic. Meaning that each commit you make contains changes that form a single unit, one should be able to check out your branch at any commit and it should still work. I see a couple of WIP commits, it's best if you squash those together. If you want some extra context please give this a read :) https://www.aleksandrhovhannisyan.com/blog/atomic-git-commits/

rcflyinghokie:
> We should only be passing breakers to systems and do any power merges there if they get fed from both main buses. Creating new powermerges outside that system will add unnecessary clutter to the LEM and Saturn code that is hard to maintain.
> 
> Can you move the Powermerge to the DSKY class and wire it in the Init function?
> 
> I also want to reinforce that it is good practice to keep your commits atomic. Meaning that each commit you make contains changes that form a single unit, one should be able to check out your branch at any commit and it should still work. I see a couple of WIP commits, it's best if you squash those together. If you want some extra context please give this a read :) https://www.aleksandrhovhannisyan.com/blog/atomic-git-commits/

I don't think I can move the power merge to the DSKY class as its needed for one set of lights in the CM and one set in the LM. Its not like its for the same set of lights as both. In other words the segment displays in the CM need the computer MNA/MNB power but only need 1 cb in the LM and visa versa.

If the LM had its own DSKY class it would be easier, but this is sharing between the vehicles currently. LCA will replace the LM power merge eventually and then the power merges can be removed.

How do you squash commits?

ThymoNL:
About squashing: If you want to make additions to a commit.
* Make your changes
* `git commit --fixup <commit hash you want to change>`
* `git commit -i --autosquash origin/Orbiter2016`

---
Why do you think you cannot move the powermerge to the DSKY class? There are multiple systems that have internal power merges.
Just pass `&GNComputerMnACircuitBraker, &GNComputerMnBCircuitBraker` or `&CDR_LTG_ANUN_DOCK_COMPNT_CB, &LTG_ANUN_DOCK_COMPNT_CB` to your init function and call WireToBuses there.

rcflyinghokie:
> About squashing: If you want to make additions to a commit.
> 
> * Make your changes
> * `git commit --fixup <commit hash you want to change>`
> * `git commit -i --autosquash origin/Orbiter2016`

What changes am I making? 


> Why do you think you cannot move the powermerge to the DSKY class? There are multiple systems that have internal power merges. Just pass `&GNComputerMnACircuitBraker, &GNComputerMnBCircuitBraker` or `&CDR_LTG_ANUN_DOCK_COMPNT_CB, &LTG_ANUN_DOCK_COMPNT_CB` to your init function and call WireToBuses there.

I think you are missing the point, its not an "OR" in this case. In the CM, the segment lights use 2 breakers, in the LM they use 1. In the CM the status lights use 1 breaker, and in the LM 2. This is why I had to use 1 e_object for each set of lights in the init and not the breakers themselves.

n7275:
The way I see it, if the power connection is two diodes to the busses in the top level schematic, then I don't think there is any problem having a power merge at the top level of the class structure. It it were something that had two power sources, but the diodes were internal to the device, than I would say put the power merge inside. I want to make some power supply classes at some point but that should be an easy upgrade to this.

rcflyinghokie:
> The way I see it, if the power connection is two diodes to the busses in the top level schematic, then I don't think there is any problem having a power merge at the top level of the class structure. It it were something that had two power sources, but the diodes were internal to the device, than I would say put the power merge inside. I want to make some power supply classes at some point but that should be an easy upgrade to this.

Exactly. And when the LCA is implemented they can be changed since the e_object will change for the LM. I don't see what the hold on this is.

n7275 (dismissed):
In that case. Approved.
Maybe we can start an issue and link it to this PR as a reminder for LCA stuff down the road.

rcflyinghokie:
> In that case. Approved. Maybe we can start an issue and link it to this PR as a reminder for LCA stuff down the road.

Added the issue and linked it.

rcflyinghokie:
@ThymoNL Still waiting on instructions on how to squash commits so this can get merged

rcflyinghokie:
@ThymoNL Finally figured out the squashing, wip commits removed

indy91 (approved):

PR #979, opened 2023-05-13
Apollo12 Checklists, MCC and scenario update

indy91:


rcflyinghokie:
If the checklist files here are from my latest update of them from the heating branch (they were up to date when you did the merge conflict resolution) then this should be good to merge, officially adding Apollo 12 support.

When this is moved out of draft I will add my review.

indy91:
Yeah that's the only downside to doing it this way, you can't see the commit history of the files anymore. But the checklists are definitely the latest from your Apollo12Checklists_MCC branch, so they should be good. I just want to check if I have missed any file on the MCC side, then I will move it out of draft.

rcflyinghokie (approved):
All of my MCC changes are present here, and the scenario file and checklists are current. I say this is ready.

PR #980, opened 2023-05-14
Reset radar interrupt when CMC wants data from the VHF ranging system…

indy91:
… but it is off or doesn't exist. Prevents CMC lockups that prevent sextant mark processing.

This is the same solution as we use for the LR and RR, because the Virtual AGC doesn't generate RADARUPT internally.


rcflyinghokie (approved):
Tested on Apollo 17 and 16, no more lockup waiting for VHF data preventing sextant marks, so behavior is correct at least.

PR #982, opened 2023-05-15
Add LM Tapemeter Power/Signal Failure light & logic

rcflyinghokie:
This adds the power/signal light functionality to the LM. This indicates a number of power/signal fail conditions assuming 2VDC is present to light the light itself:

- Tape meter loss of AC power (<85VAC) or DC power (<20VDC)
- PCMTEA 512KHz timing signal loss
- Rendezvous radar range or range rate signal loss
- Landing radar altitude or altitude rate signal loss
- PGNS altitude or altitude rate signal loss
- AGS altitude or altitude rate signal loss


rcflyinghokie:
This currently works properly for power losses and LR/RR signal failures. It is still missing range/rate data from the PGNS and AGS as well as the 512Khz timing signal from the PCMTEA, all of which are not generated currently to my knowledge.

rcflyinghokie:
PCM/TE signal added by checking cb power for now

rcflyinghokie:
All logic signals added, works in VC, also removed driving to zero logic when signal is lost making this light that much more important. Still want to add non zero/random tape initialization on LM creation.

rcflyinghokie:
This now Initializes tapemeter to nonzero value based on LM closeout images

indy91 (approved):

PR #989, opened 2023-05-19
CMVC Free Flying Viewpoint

maxq519:
Allows the CM VC to not be locked to a fixed camera position.
The concept is the same as the 2D panels, so the VC can be scrolled left/right/up/down/in/out freely.

Keys:
Arrow keys: Scroll left/right/up/down
Insert/Delete: Scroll forward/backward
Shift: Scroll slowly

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
@@ -3303,7 +3303,7 @@ int Saturn::clbkConsumeDirectKey(char *kstate)
 			if (setFreeCam == true) {
 				VCFreeCam(camDir, camSlow);
 			}
-			return 1;
+			//return 1;
 		}
 	}
 
--- a/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
@@ -3298,8 +3298,8 @@ int Saturn::clbkConsumeDirectKey(char *kstate)
 		}
 	}
 
-	if ((!KEYMOD_CONTROL(kstate)) & (!KEYMOD_ALT(kstate))) {
-		if ((oapiCockpitMode() == COCKPIT_VIRTUAL) & (oapiCameraMode() == CAM_COCKPIT)) {
+	if ((!KEYMOD_CONTROL(kstate)) && (!KEYMOD_ALT(kstate))) {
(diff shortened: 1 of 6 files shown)


jalexb88 (commented):
I just tested this and it works great, makes its much easier to move around the cockpit. 

2 Things:

1. Could the keys for forward backwards be set to shift up/down instead of insert/delete? I think that might be a bit more intuitive and my laptop actually has insert/delete merged into 1 unfortunately.

2.  Any plans to add this to the LM as well?

maxq519:
>1. Could the keys for forward backwards be set to shift up/down instead of insert/delete? I think that might be a bit more intuitive and my laptop actually has insert/delete merged into 1 unfortunately.

Sure, that's pretty easy.  However, I've always found it more intuitive to not have to press a modifier key to do things; allows for one handed operation without as much thought.  The laptop keyboard situation is a good example though.
Would it work if I made it so either Shift+Up/Down or Insert/Delete worked so the user could use whichever combination is best for them?

> 2\. Any plans to add this to the LM as well?

Should be easy enough, I'll take a look at it.

jalexb88:
Yeah that makes sense...I just looked at insert/delete on my PC keyboard and it is indeed more intuitive then I thought, but maybe having the 2 different combinations would be good to support non standard keyboards. However I just thought, is shift up/down used by anything else? If not I would vote to have both shift up/down and insert/delete.

maxq519:
> Yeah that makes sense...I just looked at insert/delete on my PC keyboard and it is indeed more intuitive then I thought, but maybe having the 2 different combinations would be good to support non standard keyboards. However I just thought, is shift up/down used by anything else? If not I would vote to have both shift up/down and insert/delete.

Yeah, I think having both is the way to go.  Insert/Delete is right above the arrow keys on full keyboard, but they're combined on my laptop as well.  I don't think Shift+Arrows is used for anything else.

maxq519:
I used Shift for fly slowly, so I went with Alt+Up/Down instead.

maxq519:
So I was using Shift for the fly slowly feature.
I decided to use ` instead of Shift for the Up/Down arrows alternate.

New keys: (Same in both CSM and LM)
Arrow keys: Scroll left/right/up/down
Insert/Delete: Scroll forward/backward
Shift: Scroll slowly
`+Up/down arrows: Alt. scroll forward/backward

jalexb88 (dismissed):

jalexb88 (dismissed):

indy91 (approved):

PR #991, opened 2023-05-22
Fix incorrect N70/71 order in P51/P52 checklists

rcflyinghokie:


rcflyinghokie:
Also adds a missing switch to LM SBD-SLEW in Apollo 10-11 LM activation checklists.

indy91 (approved):

PR #998, opened 2023-05-25
Update A8 P23 checklist to remove useless PRO

ThePuzzlemaker:
Previously, in the P23 checklist for A8 (I don't know if it's there for any others), at the end of the auto maneuver it calls to press PRO then press ENTR. But assuming the auto maneuver worked correctly, you should only need to press ENTR.


rcflyinghokie:
I see which "PRO" you are talking about here. However it can be viewed as an extra trim to attitude, so not really causing any issues.

When making a change to a procedure in a checklist, it really needs to be changed in all checklists that contain this procedure. In this case it would need to be changed in all missions as well as the other more advanced P23 checklists on later missions. Just changing it on one mission like this isn't the best approach.

rcflyinghokie (changes requested):
See comment above about procedural changes.

indy91:
@ThePuzzlemaker this pull request couldn't be merged because checklist files aren't detected as code, but as binary files. So git wouldn't have a chance to do a merge on it if the file has been changed by someone else. And as @rcflyinghokie said it's not really that much of a checklist error. So I'm closing this PR. Because of the issue with the checklist files it's probably better to ask for any checklist change on Discord, IRC or the forum. Sorry!

PR #1001, opened 2023-05-27
Allow pipes to use inlet and outlet valve size for flow calc,

n7275:
Also refactor the pipe type code a bit.


indy91 (dismissed):

n7275:
There are a few valve sizes to adjust in the LM.

For context, I'm adding this graph to the PR
image

n7275:
I added a small debugging oapiWriteLog call (nominally commented out), and was able to catch a few typos. This should be good to merge, with some cavaets to users about updating mid-mission. It might not hurt us to re-"fly" the -20 and -60 second mission scenarios, since those are probably more commonly used, and easy to recreate.

n7275:
This probably isn't a good idea. Too much is changed for too little gain in functionality and I have much better way to get the result I want for the liquidDensity PR. Namely, just creating one new valve that's the correct size and pointing the pipe I want to it. The problem that lead me down this very long rabbit-hole is that the CSM ecs has far too many, many-to-one pipe/valve configurations.

PR #1002, opened 2023-06-01
Replace Apollo 11 flightplan with one that has Revision A changes

maxq519:


rcflyinghokie (dismissed):
Looks good, just increase the install size of course since its the HSK flight plan.

maxq519:
I compressed it as much as I could without loosing too much quality.
I can compress it more and see how the quality is if you want.

-- Edit: --
I can't get the size down any more even using maximum compression settings.

indy91 (changes requested):
I'm not really sure what this does. The HSK flight plan has some, but not all the revision A changes, so we consider it the superior version? What is this change in page 1-10, changing the launch time on July 16 to 1209 EDT. That is wrong and not in the HSK version. It's the July 21 launch date that was in error, like the change document says: https://de.scribd.com/doc/46181272/Apollo-11-Final-Flight-Plan-Revision-A

Haven't checked for any more errors yet.

indy91 (approved):

PR #1008, opened 2023-06-15
RTCC coordinate system change

indy91:
The Apollo program (especially the RTCC and AGC) used a yearly changing coordinate system that had the mean Earth north pole as its z-axis. This allows the assumption that the coordinate system is very close to equatorial for the duration of the year, which leads to simplifications of many calculations, and with that a calculation speed improvement.

In the early days of our RTCC only one of these yearly changing coordinate systems had to be supported, as only Apollo 7 and 8 were flyable in NASSP, the time range being July 1st 1968 to July 1st 1969. So it was covenient to use this same coordinate system in the RTCC MFD. Many calculations like the coasting integration routine could be implemented based on AGC equations and no additional conversions had to be done. When Apollo 11 and later started being possible the question arose what to do about the coordinate system. I then decided to use the same fixed coordinate system that Orbiter itself uses, an ecliptic coordinate system, for all missions. This had (and has) many advantages, like not having to do conversions when getting data from the Orbiter API and not having to change some fundamental constants, like star unit vectors, for every yearly changing system.

Since then though we found many actual RTCC documents, from IBM and the MSC, and learned how it actually worked in much greater detail. In many cases, like the Return-to-Earth processor, calculations make use of the assumption that the coordinate system is nearly equatorial. In these cases I was never able to use the actual RTCC equations, at least not without many additional conversions. Some calculation modes were even completely impossible to do and I had to come up my own system, or postpone implementing these features. So over time the desire to change the coordinate system of our RTCC to the correct one just kept growing. The best day to do this change was always yesterday, with the RTCC always growing and changing something this fundamental just becoming more and more difficult over time.

So now I finally tackled this topic and decided to make the switch. Here is an overview of the 4 fundamental systems as we use them in NASSP right now and what they are getting changed to:

Before:

Earth-centered Inertial (ECI). Identical to the Orbiter coordinate system, but right handled (Orbiter uses left handed). The X-Y plane is in the ecliptic.
Earth-centered True (ECT). Non-inertial, rotating with the Earth. X-Y plane in the Earth's equatorial plane, Z-axis along true polar axis. X-axis pointing to prime meridian (Greenwich).
Moon-centered Inertial (MCI). This coordinate system is the same as ECI, but with the origin at the center of the Moon instead of the Earth.
Moon-centered True (MCT). Non-inertial, rotating with the Moon.X-Y plane in the Moon's equatorial plane, Z-axis along true polar axis. X-axis pointing to prime meridian.

After:

RTCC_Coordinate_Systems

from the IBM RTCC Apollo Programming Systems writeup of the coordinate system conversion subroutine.

So out of these 4 systems only MCT was previously identical to the actual RTCC. In the actual RTCC the ECT system was (quasi) inertial and didn't rotate with the Earth, so that is another relevant change this pull request does, aside from the change from ecliptic to Earth equatorial.

Looking into the future, another important thing to mention is Skylark (AGC version flown on the Skylab missions and ASTP). In both RTCC and AGC they did finally switch to a coordinate system that does not change every year, but instead uses a fixed epoch of 1950. To not have the RTCC calculations break down should we ever get Skylark I did not go "all in" on using the near equatorial assumption in some important places, like the coasting integration routines. If there are any calculation issues for Skylark remains to be seen, but it should hopefully just be one more change of coordinate system then. We could probably not support the fixed 1950 system with some planned RTE changes I want to do though. But Skylark can't fly to the Moon anyway and if we are planning to create custom AGC ropes to fly to the Moon after 1973 we should probably just switch to the yearly coordinate system of the relevant year instead of trying to implement the Skylark routines for the 1950 system in them.

So to sum it up:

Pro:
-It's the system that was used for the actual RTCC through Apollo 17
-No conversions necessary for AGC uplinks (and downlink)
-Allows more actual RTCC equations to be used without additional conversions

Contra:
-Not one fixed system across all missions, which is likely the superior way over all
-The assumptions of a nearly Earth equatorial system can make RTCC calculations break down if a mission is flown that is removed in time from its coordinate system defining epoch. This gives less flexibility with fictional missions. But those would require custom AGC ropes we haven't made yet
-Orbiter API numbers (like a state vector) have to be converted


n7275 (dismissed):

n7275:
I tried putting a few maneuvers into the MPT in Apollo 10 and 11 saved scenarios. Everything seems to calculate fine but I'm seeing HA and HP of 0.0 on the MPT (I'm relatively sure that this is me not remembering something though.) The RTCCMFD VECPOINT utility was not giving me the right attitude, not sure if that's a REFSMMAT thing.

indy91:
The issue with the HA and HP on the MPT display should be fixed now.

n7275 (approved):

PR #1013, opened 2023-06-19
Add optional LM Blast Deflectors

maxq519:
- New parameter in the mission file for no deflectors
- Descent stage now reads the mission file directly


n7275:
Hey, I think adding the ability to have the deflectors off for certain missions is a great idea. Looking through your commit, it looks like some stuff now depends on APOLLONO. It might be better to just have LMHasDeflectors 0 or 1 saved in the scenario. We used to do a bunch of stuff based on that variable in the past and had somewhat gotten away from it. Just my 2¢

indy91:
@n7275 I think it is only indirectly looking at APOLLONO. This PR makes the descent stage use the mission file, which is (right now still) being loaded based on the Apollo mission number. When the mission file is loaded then any mission specific parameter can be set in the file instead of having to be saved as an additional parameter in scenarios. The descent stage code should not use the Apollo number directly to decide if a certain configuration is used. But loading the mission file the same way as the CSM and LM are currently doing it seems fine to me.

In the long run it should not be a number, but probably the whole name of the mission file that is saved in scenarios. That's better for custom missions and also Skylab and ASTP for example, which weren't just Apollo 18+.

n7275:
Okay. Just as long as there are no if(APOLLONO == X) instances, and it only functions at metadata, then that's fine.

indy91:
Yeah I definitely would want to avoid code like that. There is still some more to get rid of in the LM, not so much anymore in the CSM.

indy91 (approved):

PR #1026, opened 2023-07-10
Advanced Visuals in Command & Lunar Module Virtual Cockpit

Jordan-64:
Improvements in CM and LM VC

- Textures with more legible text
- New meshes for the Rotarys, Switches, FDAI, DSKY.
- Added Armrests in the Lunar Module
- Generally better visuals in both VC


CaptainSwag101 (commented):
Is this code effectively the same as what's in your "HighRes-VC-Textures" branch? If so, I've been testing it for quite some time and think it's working great so far. If there's no additional changes besides what I've been testing on the old branch, I will approve.

indy91:
Functionally this is ready for merging. The only thing I still want to check is some improvements to the code. Especially the TexMul variable. I believe there are a few cases where it is used on every timestep which is really not necessary. I should get to it in the next few days and maybe will do some improvements to the code, without changing anything functionally.

n7275 (approved):

indy91 (approved):

PR #1029, opened 2023-07-15
Improvements to the RTCC moon-centered Return-to-Earth logic

indy91:
-Improve DV optimization of pre-pericynthion maneuvers
-Fix cases where previously the calculation selected a trajectory that went the wrong way around the Moon; under the right circumstances try both cases (circumlunar and non-circumlunar) for the optimum solution


maxq519 (changes requested):
In some cases, FCUA gives an undesirably large burn solution.

maxq519 (approved):
Looks good
FCUA bug remains as-is per discussion

PR #1041, opened 2023-08-01
New method to initialize the MPT

indy91:


rcflyinghokie (approved):
New buttons are much easier to use and make more sense. Only thing I cannot decide is if they should be above or below the arrow/set buttons. Below would make them more "visible" but of course it removes the MED button by the MED itself. I am indifferent here.

PR #1046, opened 2023-08-10
Reload cue cards at CM/SM sep. Move probe mesh code to AddCMMeshes function

indy91:
During the initial cue cards work I added a function called AddCMMeshes, where in the long run all the meshes related code should go that are always required in the Saturn class, so the barebones CM. A lot more can be moved there, but as part of the fix for cue cards vanishing at CM/SM sep, I am starting to do that. So this PR does:

-AddCMMeshes now handles docking probe, optics cover and cue card meshes (docking probe is new)
-AddCMMeshes is called in SetReentryMeshes, so that the cue cards are still loaded at CM/SM sep
-Duplicate code removed from SetCSMStage and SetReentryMeshes, so that it is now only done in AddCMMeshes 

In terms of changed behavior, the cue cards bug should be fixed and the optics cover can now be loaded on the CM, if it is still attached

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/saturnmesh.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/saturnmesh.cpp
@@ -1076,18 +1076,7 @@ void Saturn::SetCSMStage (VECTOR3 cg_ofs)
 	meshidx = AddMesh(hCMInt, &mesh_dir);
 	SetMeshVisibilityMode(meshidx, MESHVIS_EXTERNAL);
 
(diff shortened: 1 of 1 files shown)


n7275 (approved):

maxq519 (changes requested):
Looks like the reflective material definition for the optics cover mesh in the D3D9 config file got messed up.
Other than that, looks good.

maxq519 (approved):
Looks like something else in the past broke the reflective optics cover.
Looks good to merge this.

PR #1049, opened 2023-08-15
Upgraded S-IVB lunar impact targeting

indy91:
Improvements are:

-Add trajectory evaluation option, where the impact location of the current trajectory can be determined. If there is no impact the flyby altitude is given
-Change iteration variables so that the targeting doesn't depend on a good initial guess and converges much better
-Simplify display so that no initial guess has to be input, as long as the current trajectory impacts the Moon
-Earth impact cannot be falsely detected as a good impact
-Fix coordinate system bug in common RTCC function

Scenario I used for testing. Apollo 13 at 5h GET:

[Apollo 13 - Before TLI 0008.scn.txt](https://github.com/orbiternassp/NASSP/files/12352510/Apollo.13.-.Before.TLI.0008.scn.txt)


maxq519 (approved):
Looks good

PR #1050, opened 2023-08-17
LM optics support display in RTCC MFD

indy91:


maxq519 (approved):
Tested several cases, works great.

PR #1051, opened 2023-08-19
Saturn IB can launch on the mobile launcher

indy91:
This PR makes many changes to enable the Mobile Launcher class to work with both Saturn Vs and Saturn IBs, and in the case of Saturn IBs load a mesh with the "milkstool". The many changes and additional files are necessary to make the Tail Service Mast of the Saturn V and the Short-Cable Mast Umbilicial of the Saturn IB derived classes from a common "tail umbilical" class.

Things to do, but probably not as part of this PR, unless it is strongly requested:

-Mobile launcher mesh is a very old one we still had in NASSP, not very pretty
-No animations are yet defined for this mesh. The mesh groups are very different than our current Saturn V ML, so I might just wait until someone creates a new mesh derived from our current Saturn V ML before adding animations, then with little difference to our current animation code

Things to test:

-Saturn Vs still launching in old scenarios
-Saturn IB launching in the Skylab 2 scenario
-Service arms work and look properly


maxq519:
Looks good, but I was wondering about the launch plume effects for Saturn 1B.
There are no plumes until vehicle first movement unlike with the Saturn V, or 1B on LC-34, and they are to large and continue to long based on Skylab/ASTP missions launch video.
image

indy91:
The liftoff stream is now like in the LC-34 class for Saturn IBs

maxq519:
The plumes still look about twice as big and dense as they should based on the video.
LC-34 has the Saturn at ground level, so it will get bigger plumes than you'd get up on the milkstool.
It doesn't look like they should go much past the concrete hardstand, or higher than the milkstool deck.

indy91:
When I am talking about the liftoff stream, then I only mean the part that is going sideways. That's the ML code doing it, not the Saturn IB. I can try adjusting that a bit, probably can turn it down to as little as 20%. SSU/SSV have some elaporate code to determine how much smoke should be showing, also not based on mission time like ours still does. I'd like to change that eventually, but it goes a bit beyond this PR. So I think just turning down the level of the particle stream will be good enough.

indy91:
Liftoff stream reduced to 5% of previous, let me know what you think.

maxq519 (approved):

PR #1054, opened 2023-08-20
Add Skylab Vessel

n7275:


maxq519 (changes requested):
Could we have the included "Skylab 2 - Launch" scenario reference this vessel instead of the external addon?

n7275:
> Could we have the included "Skylab 2 - Launch" scenario reference this vessel instead of the external addon?

I don't see why not, the only downside is that our Skylab has no thrusters yet. It picks up any attitude rate before rendezvous that might make things hard. I can add some simple ones soon though.

indy91 (approved):

maxq519 (approved):

PR #1057, opened 2023-08-22
Remove as much CMC initialization code as possible

indy91:
Before this update, when the CMC gets first powered up, several things are being loaded into the CMC erasable memory: The launchpad latitude, longitude and altitude and also the initial value of the CMC clock, aligned to whatever launch time reference (TEPHEM) was loaded as part of the padload.

The addresses for the launchpad coordinates are CMC version specific and so is the base time for the time reference (TEPHEM0 in our code). So custom ropes might not get initialized properly. And to be able to support Skylark, which has a variable time reference (midnight July 1st before launch), the clock initialization code needed to be changed as well.

So this pull request changes:

-Launchpad coordinates are no longer loaded in code, but have to be part of the padload, as it was in reality
-The TEPHEM0 variable has been moved to the mission class, where all the normal ropes that NASSP uses have default parameters
-A custom TEPHEM0 can be loaded in the mission file for custom rope and if Skylark is being used. It doesn't have to be done for the nominally supported ropes
-Skylark specific clock initialization code, as TEPHEM has a different address. So this still depends on the rope name, but only Skylark is special

To test this PR any launch scenario could be loaded, after CMC powerup the CMC clock can be checked and after launch the REFSMMAT and state vector (which use the launchpad coordinate) need to be still good. I have checked this for Apollo 7 and 17

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_aux/Mission.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_aux/Mission.cpp
@@ -101,6 +101,7 @@ namespace mission {
 		bCSMHasHGA = true;
 		bCSMHasVHFRanging = true;
 		strCMCVersion = "Artemis072";
+		dTEPHEM0 = 41133.;
 		strLGCVersion = "Luminary210";
 		strAEAVersion = "FP8";
 		bInvertLMStageBit = false;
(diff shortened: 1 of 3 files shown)


CaptainSwag101 (approved):
Looks good to me, I've tested Apollo 12's Launch - 4h scenario and it seems to have a good Noun 65 that matches 24h + current GMT.

PR #1066, opened 2023-09-03
Add VHF Ranging to Skylab

n7275:
This will also clean up the CSM code a bit, and add functionality for connectors to look for each other independently of vessel-pointers (no more isLem(), etc).


n7275:
closing. this needs to be accomplished a different way.

PR #1073, opened 2023-09-07
Implement Skylab VHF ranging

indy91:


n7275 (approved):
From what I can see, VHF ranging is working as intended. The code looks great. Eventually Skylab will need an antenna with a proper gain pattern, but that's definitely out of scope.

PR #1080, opened 2023-09-13
RTCC: Fix velocity reference switch from Moon to Earth SOI in old coast integrator

indy91:


indy91:
Also fixes occultation calculation for star checks

n7275 (approved):

PR #1088, opened 2023-09-30
Checklist MFD Improvements

ThePuzzlemaker:
This PR adds a few helpful features to the Checklist MFD (N.B. cursor here refers to the currently selected step, but not necessarily the active step):
- When in a checklist, and the cursor is at the active step, an "UNDO" button is present which will undo the last completed step. If the active step is the first in the group, it will return to the previous checklist or to the "no active checklist" screen. No actions will be executed even if EXEC is enabled.
- When in a checklist, and the cursor is *not* at the active step, a "GOTO" button is present which will allow the user to skip forward or back to the step under the cursor, skipping any steps in between and not executing any actions.
- When in a checklist, and the cursor is *not* at the active step, a "BCK" button is present which will allow the user to sync the cursor with the active step. This is useful when perusing a checklist using PgDn and PgUp quite a bit.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_mfd/MFDconnector.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_mfd/MFDconnector.cpp
@@ -299,6 +299,20 @@ bool MFDConnector::completeChecklistItem(ChecklistItem* in)
 
 	return false;
 }
+bool MFDConnector::skipToChecklistItem(ChecklistItem* in) {
+	ConnectorMessage cm;
+
+	cm.destination = type;
+	cm.messageType = PanelConnector::MFD_PANEL_SKIP_TO_CHECKLIST_ITEM;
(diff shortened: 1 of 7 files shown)


ThePuzzlemaker:
I just tested this with exec enabled, and it works--i.e., it does not flip switches at all while skipping, which is the intended behaviour.

rcflyinghokie:
Tested this with a few scenarios, either I am missing something or the skip button doesn't show up at all.

maxq519 (approved):
Looks good to me

rcflyinghokie (approved):
I do like the additional functionality here, going back to previous spots is great. I would encourage documenting this a bit though explaining what the buttons do just to be safe. Especially important since orbiter MFD keyboard commands depending on which side the user has it set to will press those buttons.

But testing it the functionality looks good here.

PR #1089, opened 2023-09-30
Changed att set thumbwheel sound

TINYBOB1:
Your eardrums will thank you :)


maxq519 (approved):
Looks (sounds) good to me

CaptainSwag101 (approved):
Interesting sound, I think it'll take me some getting used to but it seems to be a little easier on the ears. I almost want to say there's a little too much reverb in the sound, but honestly I think it's an improvement regardless. Might be something we can try tweaking more in the future.

PR #1090, opened 2023-10-01
Quadrilateral click mode for VC toggle-switches

jalexb88:
VC toggle-switches are now manipulated with the left mouse button only (right mouse button only for switch covers), the coordinates of the click determine the action. This makes the switch clicking work like in the 2D panel.


CaptainSwag101:
Looks intriguing, I'll try and check this out over the coming week. I am in favor of this change sight-unseen because it will bring more parity between 2D and 3D, and will help prevent accidental clicks when attempting to use right click to pan the camera in 3D view.

CaptainSwag101:
A question about switch inputs brought up by a discussion elsewhere about this PR: 

"Can we do click-and-drag? Guessing directionality based on coordinates may be bad when the VC is shaking due to head-tracking controls being imprecise and/or programmed shaking due to thrust effects." 

This is a good point I think, and thinking about it more, the 2D cockpit has the advantage of all switches facing directly at the "camera" whereas the 3D view means that trying to click a switch up or down from a distance or at an angle might be made more difficult. I do have a tendency to "reach across the cockpit" in the CSM by simply zooming in from my existing location rather than changing viewpoints with the keys.

jalexb88:
> A question about switch inputs brought up by a discussion elsewhere about this PR:
> 
> "Can we do click-and-drag? Guessing directionality based on coordinates may be bad when the VC is shaking due to head-tracking controls being imprecise and/or programmed shaking due to thrust effects."
> 
> This is a good point I think, and thinking about it more, the 2D cockpit has the advantage of all switches facing directly at the "camera" whereas the 3D view means that trying to click a switch up or down from a distance or at an angle might be made more difficult. I do have a tendency to "reach across the cockpit" in the CSM by simply zooming in from my existing location rather than changing viewpoints with the keys.

It does sound intriguing...I think SSU/SSV does it that way? For now it will be just the change to quadrilateral click-modes which makes it work like the 2D panel...the actual code that determines how to handle the clicking is still unchanged and I simply routed the VC click coordinates to the same function that handles the 2D panel switches. If we want to change the way the clicks are handled that will be a separate project in the future as I am not sure yet about how to go about that...but if someone wants to try now, have at it. Having the quadrilateral click mode is the 1st step. 

One thing I do want to change in the near future is the click handling for rotaries in the VC, so that we can get precise control of things such as lighting rheostats and antenna controls etc...maybe have those knobs work a bit like the ALT SET for the ORDEAL.

CaptainSwag101:
I think that's a good idea too, one I know has been thought about for a while. I think it would be handy to differentiate between rheostats with a discrete number of positions, to ones which operate over a continuous, nondiscrete range of values.

n7275:
Definitely would be awesome to be able to position things like the HGA and
lighting controls, like we can with the ORDEAL altitude knob.

indy91 (approved):

PR #1092, opened 2023-10-17
S-IVB propulsive LH2 venting

indy91:
This update implements a simulation of the LH2 and LOX tank pressures of the S-IVB. Several propulsive and non-propulsive vents are implemented as well:

-LOX vent (used only on Saturn IB missions shortly after orbital insertion)
-LOX dump (was already implemented before, but now is based on tank pressure as well)
-LH2 non-propulsive vent
-LH2 continuous venting system, propulsive (Saturn V S-IVB only)

To account for the CVS thrust, both the LVDC and RTCC have been updated in their trajectory propagation. The EPO state vector uplinks by the MCC are now time tagged to a few minutes before TLI ignition. In the RTCC MFD, the page to initialize the MPT is now also used to determine if S-IVB venting is taken into account for state vector uplinks, and in the future other calculations. Further change were done to the S-IVB systems class, as the relays to control the vent valves are now all implemented. This also required updates to the scheduled events in the LVDC, the flight sequence program.

All of these changes should get some further testing. The best way is to pick a random Saturn V mission and check if an accurate TLI is achieved, if the S-IVB does its post separation vents properly and, on the later missions, if a good lunar impact can be achieved. State vector accuracy of the LVDC and AGC can als be checked. Additional testing would be for:

-Apollo 9 TD&E that is done with the propulsive venting enabled. The S-IVB is thrusting about 1 ft/s towards the CSM during the TD&E.
-Apollo 7 through LOX dump, making sure the dump DV is roughly as expected.


CaptainSwag101:
Excellent, will try to check it out over the coming weeks!

jalexb88 (dismissed):

jalexb88 (approved):

rcflyinghokie (approved):
Flew Apollo 9 with this, just took a little to get used to! #1130 has the changes for the checklist to support this PR.

PR #1093, opened 2023-10-25
SM RCS minimum impulse fire

indy91:
This PR allows the SM RCS to stay activated even with time acceleration. V49 maneuvers work without going to 1x. This is achieved by:

-On the first timestep when a thruster is fired the thrust increases  by the amount of a minimum impulse firing
-Add logic for firing the RCS if the CMC commands a thruster on and has already commanded it off on the same timestep

I don't necessarily intend this PR to be merged, it's more of an experiment. I'm adding it as a draft PR so it is more visible.


MrResiduals:
A timewarp of 8 hours with the LM (100x and Low Bit Rate), results in about 5-10% RCS Usage. Much better compared to before, but I will need to do timewarp testing with the CSM only.

MrResiduals:
RCS in P20 Auto Orb Rate (With LM attached/CSM only), handled much better than my last post a month ago. Very little oscillation only. Some minor ones on the far side. Will need a little more testing regarding V49 and holding attitude before I can say this is ready.

maxq519:
It also looks like there's a debug string left over in here somewhere

maxq519 (approved):

PR #1096, opened 2023-11-03
Apollo 9 MCC: Time tag of state vector uplinked to CMC before rendezv…

indy91:
…ous is 90h instead of 92h. Too far into the future runs into AGC software limitations.


CaptainSwag101 (approved):
This fixes the long integration time and erroneous V82 display, everything seems to work as expected now.

PR #1098, opened 2023-11-08
Fix Skylab Visual Studio project not building in debug mode

indy91:


ThymoNL:
The MultiByte character set is an option intended for maintaining legacy code that needs to work with Windows 9X.
It should not be used for modern code. It wraps all Windows API functions that use a char * with a helper that translates it to wchar_t. If using Unicode breaks your build it points to a coding error somewhere and not a project setup issue.

See https://stackoverflow.com/questions/9349342/about-the-character-set-option-in-visual-studio

indy91:
The specific build error I was trying to fix was this:

Error	LNK2019	unresolved external symbol "__declspec(dllimport) public: char * __thiscall VESSEL::GetClassNameW(void)const " (__imp_?GetClassNameW@VESSEL@@QBEPADXZ) referenced in function "class Connector * __cdecl GetVesselConnector(class VESSEL *,int,enum ConnectorType)" (?GetVesselConnector@@YAPAVConnector@@PAVVESSEL@@HW4ConnectorType@@@Z)	

All of our Visual Studio projects, release and debug mode, use the "Multi-Byte Character Set" at the moment. It was just missed in the new Skylab project, in debug mode only. If you know how to fix the build error, feel free to change all our projects to using Unicode character set.

EDIT: The issue might be the line:

char *classname = v->GetClassName();

in Connector *GetVesselConnector(VESSEL *v, int port, ConnectorType t) in connector.cpp. Changing it to use GetClassNameW directly doesn't work either. So not sure how to get it to work with Unicode character set.

ThymoNL:
Hmm, okay. If we still need it for legacy reasons then I won't object. If someone feels like figuring out what exactly needs to be done so we can switch to Unicode a change would be welcome. Multi-Bye is fine until then or when it becomes problematic.

ThymoNL (approved):

n7275:
I believe the cause of this is the "BasicExcel" code that we use for
reading checklist files. It's quite old, and the checklists definitely do
not support unicorn.

indy91:
> Hmm, okay. If we still need it for legacy reasons then I won't object. If someone feels like figuring out what exactly needs to be done so we can switch to Unicode a change would be welcome. Multi-Bye is fine until then or when it becomes problematic.

Yeah, I think there are a bunch of things with how NASSP is building in Visual Studio right now that could eventually be problematic... could definitely use some upgrades by someone who knows what they are doing, so not really me.

PR #1101, opened 2023-11-13
Fix Apollo 15 specific cue card locations

indy91:
image


CaptainSwag101 (approved):
Tested on my current Apollo 15 mission, all the cue cards by the DSKY look good and don't clip any more.

PR #1107, opened 2023-11-15
Give coordinate system epoch as parameter to coasting integrator; pre…

indy91:
…vents inaccurate trajectory propagation in some cases. Use newer trajectory propagator for LOI/MCC maneuver transfer


CaptainSwag101 (approved):
Fixed my LOI maneuver computations early into the mission (flight day 3), PAD looks good now.

PR #1116, opened 2023-11-27
Mission file loading is less strict about empty spaces

indy91:
Makes the mission files a bit more user error tolerant. For example:

`HasLMProgramer=TRUE`

was the only way with the previous behavior that loaded correctly.

`HasLMProgramer=TRUE ` with an empty space at the end and `HasLMProgramer= TRUE` with an empty space before TRUE wouldn't have loaded as true.

Similar, any AGC/AEA software being loaded like `LGCVersion=Luminary099` was the only way it loaded correctly the `Luminary099.bin` file. `LGCVersion=Luminary099 ` would have tried to load `Luminary099 .bin` due to the empty space.

The downside of this, there can't be an empty space within the name of the rope binary file.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_aux/Mission.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_aux/Mission.cpp
@@ -184,7 +184,7 @@ namespace mission {
 				sscanf(line + 12, "%d", &iSMJCVersion);
 			}
 			else if (!_strnicmp(line, "JMission=", 9)) {
-				strncpy(buffer, line + 9, 255);
+				sscanf(line + 9, "%s", buffer);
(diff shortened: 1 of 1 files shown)


ThymoNL:
This is unsafe. sscanf does not know the length of the buffer you are writing to. So if a line that's too long is read in, a buffer overflow will occur.
I you want to use strings with sscanf you must specify the field with in your conversion specifier. E.g. `%255s`.

You correctly state that we lose support for values containing white space this way. It would be nice if we don't limit ourselves by removing support for it. How about we still use strncpy but trim leading and trailing whitespace instead?

indy91:
Well I guess this update is just making things worse then. I don't like the idea of a bunch of special code for getting rid of the whitespace when using strncpy either, so we just need to stay careful not to do something wrong in the mission files.

PR #1119, opened 2023-12-02
Add an HSystems class for simulating pressure/temperature of the atmosphere outside of vessels.

n7275:
This is intended as an improvement over the h_Vent class that currently only simulates venting to vacuum, and uses multiple independent volumes.

The new class: h_ExteriorEnviormnent allows multiple objects to connect to a single "venting enviormnent"


rcflyinghokie:
Works properly as a separate vent, tested with venting with battery vents

rcflyinghokie:
I think this can be merged safely as it is a stepping stone to replacing vents already in place as necessary

rcflyinghokie (dismissed):

rcflyinghokie (dismissed):

indy91 (approved):

PR #1122, opened 2023-12-03
Gunter Wendt ref

343GuiltySpark-04:
Added An engraving that says "Gunter Wendt Was Here!" to the side hatch


rcflyinghokie:
While this is a fun easter egg, I don't really see any benefit for it to be added to the main repo.

ThymoNL:
Agreed. If this engraving was not on the actual ship it won't qualify to go into NASSP.

PR #1133, opened 2023-12-29
Make the LM visible in the CSM optics from up to 400nm

maxq519:
Capture

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_lm/LEM.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_lm/LEM.cpp
@@ -549,6 +549,8 @@ void LEM::Init()
 	VcInfoActive = false;
 	VcInfoEnabled = false;
 
+	visibilitySize = 31.1; //Tuned so the LM disappears in the CSM optics at 400nm range
+
 	Crewed = true;
 	AutoSlow = false;
 
@@ -2009,6 +2011,36 @@ void LEM::clbkDockEvent(int dock, OBJHANDLE connected)
 			UndockConnectors(dock);
 			CreateAirfoils();
 		}
+	}
+}
+
+void LEM::clbkFocusChanged(bool getfocus, OBJHANDLE hNewVessel, OBJHANDLE hOldVessel)
+{
+	OBJHANDLE hLM = GetHandle();
+	if (hNewVessel == hLM) { //LM gains focus
+
+		bool fixCamera = false;
+		if (oapiCameraInternal() == false) {
+			fixCamera = true;
+			oapiCameraAttach(hLM, 0);
+		}
+		
+		if (status == 0) { //Dock Stage
+			SetSize(6);
+		}
+		else if (status == 1) { //Hover Stage
+			SetSize(7);
+		}
+		else if (status == 2) { //Ascent Hover Stage
+			SetSize(5);
+		}
+
+		if (fixCamera == true) {
+			oapiCameraAttach(hLM, 1);
+		}
+	}
+	else if (hOldVessel == hLM) { //LM loses focus
+		SetSize(visibilitySize);
 	}
 }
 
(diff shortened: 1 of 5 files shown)


rcflyinghokie:
I haven't tried this yet, but does this make the tracking light visible at that distance and the vessel marker?

rcflyinghokie:
Also, this also needs to be done for the SIVB for Apollo 7

maxq519:
> I haven't tried this yet, but does this make the tracking light visible at that distance and the vessel marker?

Yes it does.


> Also, this also needs to be done for the SIVB for Apollo 7

Easy enough

rcflyinghokie:
Works for me on the LM. Is there any performance or other hit for using 400nmi versus something smaller like 200nmi?

maxq519:
No performance hit unless you’re actively looking at the LM through the SXT.

maxq519:
Done and working for the S-IVB as well

n7275:
@indy91 and I discussed some implications of this last (2023) January. https://nassp.space/irclogs/2022-12-27-12%3A46_Log.log search for "SetSize()"

maxq519:
> @indy91 and I discussed some implications of this last (2023) January. https://nassp.space/irclogs/2022-12-27-12%3A46_Log.log search for "SetSize()"

Looks like the only thing actually affected is radiation force then?  Or did I miss something.
Overall, I think that may be a fair trade, given that flying an accurate CSM rescue is not really possible right now.

n7275:
> > @indy91 and I discussed some implications of this last (2023) January. https://nassp.space/irclogs/2022-12-27-12%3A46_Log.log search for "SetSize()"
> 
> Looks like the only thing actually affected is radiation force then? Or did I miss something. Overall, I think that may be a fair trade, given that flying an accurate CSM rescue is not really possible right now.

Yeah, I think it's probably worth the tradeoff, just wanted to make sure we discussed it.

I think the long-term solution, should be to push for a replacement for the angular spot-size, single pixel calculation. And replace it with apparent magnitude, this report has some examples: https://apps.dtic.mil/sti/tr/pdf/AD0785380.pdf I'm pretty sure this wouldn't be hard to add to Orbiter, even the F(Φ), could be set in clbkSetClassCaps.

For now, I'm pretty sure this is okay.

n7275 (approved):

rcflyinghokie:
So why would this make a CSM rescue not work?

maxq519:
> So why would this make a CSM rescue not work?

Because you can’t take optics marks until the LM is within 60 miles of the CSM right now.  It’s not that it can’t be done, just that it’s not accurate.

rcflyinghokie:
> > So why would this make a CSM rescue not work?
> 
> Because you can’t take optics marks until the LM is within 60 miles of the CSM right now. It’s not that it can’t be done, just that it’s not accurate.

I am confused, I am asking why this PR would make a CSM rescue not work

maxq519:
> > > So why would this make a CSM rescue not work?
> > 
> > 
> > Because you can’t take optics marks until the LM is within 60 miles of the CSM right now. It’s not that it can’t be done, just that it’s not accurate.
> 
> I am confused, I am asking why this PR would make a CSM rescue not work

Haha sorry for the confusion.  This PR allows a CSM rescue to work fully, prior to this PR it would not.
I was just making the point that it was a fair trade for the radiation pressure thing.

rcflyinghokie:
> > > > So why would this make a CSM rescue not work?
> > > 
> > > 
> > > Because you can’t take optics marks until the LM is within 60 miles of the CSM right now. It’s not that it can’t be done, just that it’s not accurate.
> > 
> > 
> > I am confused, I am asking why this PR would make a CSM rescue not work
> 
> Haha sorry for the confusion. This PR allows a CSM rescue to work fully, prior to this PR it would not. I was just making the point that it was a fair trade for the radiation pressure thing.

Ohh! Ok I am tracking now haha

rcflyinghokie (approved):
Works as intended

indy91:
Yeah I guess it's the best solution we can get for now in Orbiter. The size isn't excessive, so it shouldn't cause any bad side effects.

PR #1135, opened 2023-12-30
Move LR class to its own files

indy91:


rcflyinghokie:
good idea, there are probably other things that need their own files as well

rcflyinghokie (approved):

PR #1141, opened 2024-01-06
Implement options 1 to 6 of the Apollo Generalized Optics Program

indy91:


n7275 (dismissed):

maxq519 (changes requested):
All functions work great, however some of the columns on the output page may be labelled incorrectly.
For Option 1, the angles are labelled PYR, but the angles are actually RPY.
Seems like the same thing for the IGA/MGA/OGA outputs on the other pages.

maxq519 (dismissed):

maxq519 (approved):

PR #1143, opened 2024-01-06
Fix LM radiation pressure bug

maxq519:


rcflyinghokie (dismissed):

rcflyinghokie:
I put an approve in, but the SIVB without focus would also be impacted by this

indy91 (approved):

PR #1163, opened 2024-01-29
yaAGC: Implement internally-timed RADARUPTs

indy91:
Co-authored-by: thewonderidiot

Changes behavior of all radars (RR, LR, VHF ranging). It will fix instances where the radar interrupt previously happened too late (1210 alarms caused by the LR spurious routine) and when it happens too early (Skylark R27).

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_sys/apolloguidance.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_sys/apolloguidance.cpp
@@ -323,6 +323,18 @@ typedef union
 	unsigned long word;
 } AGCState;
 
+//
+// Structure for additional yaAGC bits
+//
+
+typedef union
+{
+	struct {
+		unsigned RadarGateCounter:4;
+	} u;
+	unsigned long word;
+} AGCState2;
+
 void ApolloGuidance::SaveState(FILEHANDLE scn)
 
 {
@@ -371,6 +383,13 @@ void ApolloGuidance::SaveState(FILEHANDLE scn)
 	state.u.Trap32 = vagc.Trap32;
 
 	oapiWriteScenario_int(scn, "STATE", state.word);
+
+	AGCState2 state2;
+
+	state2.word = 0;
+	state2.u.RadarGateCounter = vagc.RadarGateCounter;
+
+	oapiWriteScenario_int(scn, "ADDSTATE", state2.word);
 
 	//
 	// Write out any non-zero EMEM state.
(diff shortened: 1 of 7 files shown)


thewonderidiot (dismissed):

Orbitersdk/samples/ProjectApollo/src_sys/apolloguidance.cpp
```diff
@@ -542,6 +561,12 @@ void ApolloGuidance::LoadState(FILEHANDLE scn)
 			vagc.Trap31B = state.u.Trap31B;
 			vagc.Trap32 = state.u.Trap32;
 		}
+		else if (!strnicmp(line, "ADDSTATE", 8)) {
```
ThymoNL: Why `ADDSTATE` exactly?

ThymoNL:
So we finally ran out of space in AGState...

Looks good! Just curious about the naming of the save state.

Orbitersdk/samples/ProjectApollo/src_sys/apolloguidance.cpp
thewonderidiot: Took me a second too. My guess is it's short for "Additional State".

thewonderidiot: ...and something like "STATE2" can't easily work because of how the `scanf()` reading is set up. `STATE2` would be matched by the preceding `else if (!strnicmp (line, "STATE", 5)) {` check.

indy91: @ThymoNL  Haha yeah, Mike got it right. It's ADDSTATE for additional AGC state. STATE2 can work if functions are used that check on the whole string, which we do in other save/load functions but not here yet. So I couldn't be bothered rewriting the loading function enough so that I could use STATE and STATE2. If you have a better idea for a name, now we can still change it. Once this gets merged we shouldn't do that anymore for backwards compatibility.

ThymoNL: I get it. Using this name does limit us to only this extra state though. I expect this will be enough for quite a while, but surely that was also the case when AGCSTATE was made initially. How would we add a new state when this one runs out? You'd have to get creative with naming that one, or allow numbers to be used.

indy91: Alright, I will change it to STATE and STATE2. Just have to rework the loading of those a bit, hopefully without breaking everything!

indy91: In my testing it looks good, but I am always a bit hesitant with int to unsigned int conversions, which can't be avoided here.

else if (papiReadScenario_int(line, "STATE", inttemp))

inttemp is an integer, because that's what papiReadScenario_int reads.

state.word = inttemp;

word is unsigned long.

I checked with putting -1 as STATE in the scenario and it reads (as I believe) correctly as all bits of STATE being set to 1. Any objection to this solution?

ThymoNL (approved):

PR #1167, opened 2024-02-03
Update crew heat and water vapor for more accurate human physiology

n7275:


rcflyinghokie:
testing now with A16

rcflyinghokie:
Updating, so far this looks good in both spacecraft, will fly to Entry to verify. Only side effect is more water generation in the waste tanks, so this means wastewater fills faster in the CM and less water is used for cooling in the LM

n7275:
we probably should update water consumption to match at some point

rcflyinghokie:
Yeah I think we should add it similar to crew O2 maybe, based on crew in cabin. Ideally I would like maybe a "ration" option for flying 13 :P

rcflyinghokie:
About a third through TEC, I think we need to add water consumption based on crew to the potable tank to offset the waste tank filling so fast from potable overflow

rcflyinghokie:
I think this is safe to merge as is, once merged I will fix merge conflicts with https://github.com/orbiternassp/NASSP/pull/1186 so we can have better water balance

rcflyinghokie (approved):
See previous comments

PR #1168, opened 2024-02-03
PAMFD and Checklist MFD show RTCC ground elapsed time

indy91:


rcflyinghokie:
Works correctly in PAMFD, need to decide if to add the same logic to checklist MFD before releasing or down the road.

indy91:
I'm happy with this now.

rcflyinghokie (approved):

PR #1170, opened 2024-02-04
Transition from OrbiterSound to XRSound

CaptainSwag101:
This is aimed to move to a sound engine that is less prone to causing random crashes, and to ease the path towards OpenOrbiter compatibility. This implements vessel-class-specific config files to disable most default sound groups like cabin ambience, ATC, and other generic callouts, while not clobbering any defaults if the user has other addons installed besides NASSP.

It also adds tweaked versions of XRSound's default SFX in a couple places to make them a little more pleasing to the ears, such as reducing the volume of the RCS mode callouts "rotation" and "translation", as well as applying a filter to the Main Engine sound loop to make it less crackly.

Besides these changes, most of the work should be transparent to the end user and the developer, except where certain differences in implementation require changes (such as a fix for switch/button sounds not playing when loading into the 3D cockpit without viewing the 2D cockpit; a one-line fix.)


CaptainSwag101:
I'm double-checking an issue where our modified main engine sound isn't being respected. I think it may have to do with how we apply the sound manually since the thruster group doesn't belong to "main engines" but is instead the "user" group.

rcflyinghokie:
How is this looking as far as being pushed? I have had no issues with it thus far.

CaptainSwag101:
> How is this looking as far as being pushed? I have had no issues with it thus far.

Main thing I'm trying to do is see if I can fix the issue where high frame rates break fading sound effects. Unfortunately it may require some rework of the SoundLib intermediary code that interfaces between NASSP and XRSound (or previously OrbiterSound). It currently represents volume as an integer between 0 and 255, which I think is where the trouble arises. Might be best to convert to a float/double between 0.0 and 1.0.

CaptainSwag101:
Okay, with the high-framerate fade bug finally squashed (as far as I can tell), I believe this PR is ready for final testing and prime-time!

indy91 (approved):

PR #1179, opened 2024-02-13
Remove LM preg cabin condition for repress engaging/master alarm

rcflyinghokie:


rcflyinghokie:
Testing this a bit further but seems to be the correct behavior, will release from draft after final checks. Code will remain commented out for the time being.

rcflyinghokie:
Behavior matches procedures and transcripts, safe to merge.

indy91 (commented):
Shouldn't the variable "repressinhibit" be completely removed with this change?

rcflyinghokie:
> Shouldn't the variable "repressinhibit" be completely removed with this change?

Yep I can comment it out for now as I did the other code.

rcflyinghokie:
After more investigation, this changed from LM-6 to LM-7, so the old behavior is still applicable for LM-6 and previous.

indy91 (approved):

PR #1184, opened 2024-02-18
Panel 252 Typo Fix

GoForPDI:
Change Waste Stowage Vent from 'stoage' to 'stowage'. This does not affect the VC, and the font is not the right one, it is what looked closest from the options Paint gave me. So this will need to be edited in the future, but it's the best I can do at the moment.


GoForPDI:
Filesize is unexpectedly ballooning so not ready for merge yet.

rcflyinghokie:
is this the same as https://github.com/orbiternassp/NASSP/pull/1220

PR #1189, opened 2024-02-25
CMC translation commands work in Accel Cmd

indy91:
With the previous behavior CMC translation commands were inhibited if the Manual Attitude switch for an axis was set to Accel Cmd. SCS translations weren't affected.

This behavior is wrong as the RJEC switches to an (Accel Cmd only) SCS mode on an axis per axis basis. Any translation commands for the specific axis are then actually processed through the SCS electronics and aren't coming from the CMC anymore. That way, even with S/C control set to CMC, translations still work.


rcflyinghokie (approved):
Behavior matches the procedures now, no other ill effects noted

PR #1190, opened 2024-02-26
Add working window shades in the LM VC

maxq519:
Mesh and texture by @TINYBOB1


Orbitersdk/samples/ProjectApollo/src_lm/LEM.h
```diff
@@ -1657,6 +1664,7 @@ class LEM : public Payload, public PanelSwitchListener {
 	UINT ascidx;
 	UINT dscidx;
 	UINT vcidx;
+	UINT windowshadesidx;
```
indy91, line 1667: windowshadesidx should be initialized to -1 in LEM.cpp like the other indices.

Textures/ProjectApollo/VC/AditionalParts.dds
indy91, line 1: What's the reason for the file size increase?

maxq519: Ah, must have exported it as DXT3 instead of 1.  Will fix that

indy91 (approved):

PR #1191, opened 2024-02-27
Save splashdown coordinates in RTCC class instead of RTCC MFD

indy91:


rcflyinghokie (approved):
Confirmed coordinates did indeed save

PR #1194, opened 2024-03-05
RTE ATP target lines and PTP target coordinates can be changed in the RTCC MFD and are saved/loaded in scenarios

indy91:


rcflyinghokie (approved):
Functions as intended, tested with Apollo 16 launch date change

PR #1198, opened 2024-03-10
RTCC GMT was zero during scenario load, causing trajectory updates to…

indy91:
… fail; add error checking if anchor vector for trajectory update could be moved to present GMT or if it had instead stopped propagation at atmospheric entry

The attached scenario fails to load because pull request #1168 introduced a RTCC GMT elapsed time variable that is zero on scenario load. This causes the trajectory update during scenario loading, which happens if a MPT trajectory update had been done previously, to propagate the state vector to 0h GMT instead of present time. In the scenario this caused the state vector to impact the Earth atmosphere. This state vector was then used to generate the ephemeris, which failed immediately due to being in the atmosphere.

Aside from fixing the zero GMT bug this update also adds additional error conditions for the atmospheric reentry case.

[Apollo_16_Littrow_75_50_00.scn.txt](https://github.com/orbiternassp/NASSP/files/14550333/Apollo_16_Littrow_75_50_00.scn.txt)


rcflyinghokie (approved):
loads the frozen scenarios now

PR #1199, opened 2024-03-10
Update appveyor.yml to package XRSound configuration separately

dseagrav:
This is so people can change xrsound's configuration but still use zipfiles for updates


CaptainSwag101 (approved):
Looks good to me.

CaptainSwag101 (changes requested):
Could the new zip file be called "XRSound-config" perhaps instead of just "XRSound"? A user had some confusion and thought this was XRSound itself when it's just the default configs.

indy91 (approved):

CaptainSwag101 (approved):
Everything looks good now, thanks!

PR #1201, opened 2024-03-15
Implement new version of the time-theta function; fixes rare hyperbolic error cases

indy91:
In some rare cases the commonly used function time_theta returned wrong results. To really fix this the function had to be rewritten from scratch. The specific case that failed happened in the attached scenario. To reproduce it go to the lunar map update page and enter the time 78:50:00 GET. The scenario is before LOI, but this input time is far enough beyond pericynthion that it triggered the time_theta bug. 

Note that the calculation for the prime meridian crossing time (PM) prefers a near circular orbit, so that with this input time the correct crossing time can not be found even with this update. But error checking is in place to prevent a stuck calculation or CTD. The actual issue in the scenario was with the sunset calculation which used the wrong time_theta output in an iteration loop that then resulted in a sunset time so large that the MFD caused a crash due to overflow.

[A15_44_Pre-LOI_SV_Update.scn.txt](https://github.com/orbiternassp/NASSP/files/14618441/A15_44_Pre-LOI_SV_Update.scn.txt)

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_launch/rtcc.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_launch/rtcc.cpp
@@ -3307,7 +3307,10 @@ RTCC_PMSTICN_3_1:
 	}
 	if (T1 >= T2)
 	{
-		T2 = T1 + OrbMech::time_theta(opt.sv_P.R, opt.sv_P.V, WT, mu);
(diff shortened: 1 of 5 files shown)


maxq519 (commented):
Looks good to me

maxq519 (approved):

PR #1203, opened 2024-03-31
Add missing Apollo 12 MCC updates

indy91:
This is mostly all of the lunar orbit map updates, but also a few photo PADs. Because there wasn't enough space for MCC updates in between, this update will break old Apollo 12 scenarios after TEI. This update also adds an option to the RTCC MFD to use 150°W or 180°W as the prime meridian for lunar orbit map updates.


rcflyinghokie:
I will review this along with the checklist to make sure everything syncs up

rcflyinghokie:
Added a PR with some final times

rcflyinghokie (approved):

PR #1206, opened 2024-04-05
RTCC Guide - MSK 1591 => 1590 for Vector Comparison Display

ThePuzzlemaker:


ThePuzzlemaker:
Looks like GitHub automatically converted line endings. If you want me to reopen with CRLF line endings, let me know.

indy91:
Also, we have no automatic PDF file building set up, so any change to the manual .tex file should also include a change to the PDF.

ThePuzzlemaker:
Got it, I'll rebuild.

ThePuzzlemaker:
Hopefully I did that right :sweat_smile:
The LaTeX build process has always confused me, even with latexmk

indy91:
> Looks like GitHub automatically converted line endings. If you want me to reopen with CRLF line endings, let me know.

I don't know about reopening, but this is a potential issue. Is that something you can fix on your end so that the .tex file doesn't have a diff that includes all lines?

ThePuzzlemaker:
Think that fixed it, the diff still looks a little weird but I think the line endings aren't mixed anymore

ThePuzzlemaker:
It's at least a manageable diff now

indy91:
Yeah I have had mystery line changes sometimes as well, don't think it's the line endings. Thanks for fixing the diff!

indy91 (approved):

Issue #1209, opened 2024-04-11
Virtual Cockpit GPI trim wheels dont save position/round when scenario reloaded


rcflyinghokie:
As the title states, its a known issue that the trim wheels in VC do not save properly to the .1 place when a scenario is reloaded.

indy91:
Fixed by #1213

PR #1213, opened 2024-04-19
New classes for rotational switches and thumbwheels with finer controls

indy91:
A new type of switch class has been added that has a double precision state instead of integer, discrete positions. This class is used for the SPS gimbal trim thumbwheels, the ORDEAL altitude selection and all the lighting rotational switch controls in CSM and LM. Also the CSM high gain antenna and LM steerable S-band antenna pitch and yaw controls. The precise state of these switches is only visible in the virtual cockpit, as the 2D panel has a limited number of bitmap to represent the states of the switches.

Testing of this branch would be to check if all these switches still behavior correctly in 2D panel and Virtual Cockpit.


rcflyinghokie:
Working very smooth in VC currently.

indy91:
Aside from Checklist MFD changes for the CSM HGA and LM steerable antenna as well as the SPS trim gimbal thumbwheels this pull request is ready for testing.

maxq519 (approved):

jalexb88 (approved):

indy91:
This also fixes #1209

PR #1220, opened 2024-05-11
Correct text from STOAGE to STOWAGE on 2D panel 252

maxq519:
The issue with filesize inflating for no reason is still with us, the texture is an 8 bit bmp as stated by Jordan.


rcflyinghokie:
So is there any way to avoid it being 2mb larger than before?

maxq519:
Not that I know of, @Jordan-64 said it was because the original is compressed in some way.
I tried all the tricks I could think of but unless we can figure out how the original is compressed I don't think there's any way to avoid it.

rcflyinghokie:
also is this meant to replace https://github.com/orbiternassp/NASSP/pull/1184

rcflyinghokie (approved):
Works in 2D and VC

PR #1222, opened 2024-05-13
Add flip flop quirk to LM heater caution

rcflyinghokie:
Based on the LM pilots handbook, the capped off LR caution (we assume LM-5 and subs) causes a flip flop quirk on CWEA powerup. This has been replicated. For later missions, the same quirk is present for capped off RCS quad caution FFs (LM-7 and subs.) Caps were also added for the RCS heater caution.

This also adds mission specific numbers for CWEA changes in the future.

image

Do not be alarmed if you get a heater caution light now when loading, just clear the warning with the temperature monitor rotary each time you cycle the CWEA.


rcflyinghokie:
RR capping off will have to be added later for LM-7 and subsequent, will not be part of this PR

maxq519 (approved):

PR #1226, opened 2024-05-15
Use only one mission file parameter with the LM vehicle number instea…

indy91:
…d of having separate CWEA and EPS/ECS systems paramreters


rcflyinghokie (approved):
Tested behavior on LM-3 LM-4 and LM-6 all work correctly

PR #1227, opened 2024-05-16
Add control in Orbiter launchpad to adjust the level of Virtual Cockpit vibrations

indy91:
image


CaptainSwag101 (approved):
Works well for me, and the best part (in my opinion) about this implementation is that you can go over 100% intensity if you want some extra hilariousness! You can also set the value negative if you want, which works as you'd expect: inverts the magnitude of the effects. It should be noted that this affects all camera-related movement in the capsule, including CM RCS firing "shakes" and the zoom-out caused by entry G-load, etc.

PR #1230, opened 2024-05-18
Replace "Braker" with "Breaker"

rcflyinghokie:
I have been wanting to do this for a while and finally made a PR. Please check this! It builds and seems to work properly but as with any find/replace there is still potentially error.


rcflyinghokie:
making this a draft as all old scns have to be fixed for it

PR #1242, opened 2024-05-27
EASEP/ALSEP Experiments

TINYBOB1:
Adds the EASEP experiments and equipment for Apollo 11.  This includes the TV camera, the Solar Wind Composition Experiment, the Passive Seismometer Experiment, and the Lunar Ranging Retro Reflector.  Deployed as per the flightplan order with the "K" key.  CDR does the camera, and LMP does the rest.  Thanks to @maxq519 for the code to get them to deploy!
image
image
image


rcflyinghokie:
Is this locked to Apollo 11?

TINYBOB1:
I believe so, everything but the TV camera, which is set to 14 and below, but 12-14 used a different camera then 11.  I can make a mesh for it and see if I can get it to work if you want.  technically the LR3 should be on 14 and 15, but 15 used a different style one.

rcflyinghokie (dismissed):
Tested on 11, only suggestion is on release make a post about it in the release work thread with instructions.

TINYBOB1:
the solution file has a conflict, not sure how exactly to fix, since both have elements that need to be taken.

indy91 (approved):

rcflyinghokie (approved):
Adding back my review, works as advertised!

PR #1246, opened 2024-06-05
Clean up Apollo 11 CSM Maneuver PAD calculations

indy91:
Makes CSM Maneuver PAD calculations more "state of the art" in terms of our RTCC:

-Always uses an input state vector and doesn't calculate one itself with the help of a vessel pointer. This will allow better MPT compatibility of Maneuver PADs in the future
-Together with a state vector type that doesn't automatically contain a mass the Maneuver PAD calculation now takes a weight table with separate CSM and LM weights as input. This is a cleaner solution for taking overall weights into account
-It now uses the new RTCC propagator with drag taken into account

Testing for this PR would involve checking Maneuver PAD calculation in the RTCC MFD and some random samples from MCC missions, as ALL CSM Maneuver PAD calculations on Apollo 8, 10-12 had to be changed for this


indy91:
Probably becomes part of a bigger RTCC MFD update.

PR #1250, opened 2024-06-08
Make RTCC MFD independent of current focus vessel

indy91:
Also includes improvements of the inputs to all Maneuver PADs, which made changes necessary for all missions in the MCC mission support, which will need to be tested along with the MFD changes.


n7275 (dismissed):

ThatRedMelon (changes requested):
For just the RTCC side I've tested, it mostly seems to function well, but a few issues I've seen:

There's a few places where the display only allows input for vector time, and doesn't allow you to change the vessel, such as the LLWP and perigee adjust inputs.

Similarly, I found that the de-orbit display didn't have a vessel input, and I'm not sure if it's needed or not.

Final thing I noticed was the lunar ascent processor only had a CSM selection, and not an LM selection. I assume the lunar ascent processor take the LLWP/LLTP results for the calculations, hence the LM vessel is not needed there as its selected in other displays? If so, that works fine, just making sure.

Otherwise, tested many calculations from the MCC vessel and they all seemed to calculate correctly and give the correct results, so works really well.

(Edit: all issues accounted for)

ThatRedMelon (approved):

PR #1254, opened 2024-06-19
All event timers continue to count down when reaching zero; LM event …

indy91:
…timers from Apollo 13 on reverse direction and start to count up


rcflyinghokie (approved):
As discussed, the behavior likely needs to be fixed in the future for the count down/count up to prevent delay but it works pretty well for now

PR #1260, opened 2024-06-26
Save/load state of CSM O2 hose

indy91:


rcflyinghokie (approved):
Save load tests show it's working

PR #1267, opened 2024-07-11
Add DSKY UDP API to interface with external programs

Pablo-Ortiz-Lopez:
This PR aims to enable the possibility to get data in and out of the DSKY, using an UDP socket, in JSON format, so that it can interact with external programs which aim to emulate the DSKY. The functionality is disabled by default.


Pablo-Ortiz-Lopez:
Just tested it on OpenOrbiter and OrbiterBeta r90, seems to work in both. To test it you have two options:
### Using our DSKY software
- Enable the socket in the DSKYOut.cfg file
- Clone https://github.com/Apollo-Simulation-Peripheral-Lab/DSKY.git
- Have NodeJS installed
- Run start.bat, wait for the browser to show the dsky
- Select “no serial output” and “NASSP” in the terminal
- Run nassp (if you use true fullscreen it might crash when switching windows so i use “with taskbar”)
- When the scn is loaded you should be able to control the dsky from the web browser (verb=v, noun=n,enter=e,…)

### Using a third-party UDP testing program
- Download the program from: https://github.com/danielnilsson9/tcp-udp-tool/releases/tag/1.1.0
- Enable the socket in the DSKYOut.cfg file
- Run nassp (if you use true fullscreen it might crash when switching windows so i use “with taskbar”)
- Open the test program, go to the UDP tab, and start listening on port 3002.
- To send keypresses, select the port that the messages are being received from. The port will change if you change vessels. This lets you control which dsky you type in. The messages will be sent only from the dsky of the currently focused vessel, so the last received port will be the one to write to in order to type into the focused dsky

rcflyinghokie:
I would make this a draft until ready for review

Pablo-Ortiz-Lopez:
Marked as ready for review following a discussion on the discord server

rcflyinghokie (commented):
DSKY lighting wiring/power looks correct for the 3 DSKY's

indy91 (approved):

PR #1271, opened 2024-07-17
yaAGC: Differentiate between real TC/TCFs and transients for TC Trap

thewonderidiot:
This matches virtualagc/virtualagc#1231 -- see that PR for details.


CaptainSwag101 (approved):
Everything looks good to me!

PR #1272, opened 2024-07-17
LVDC: Treat the planned S-II center engine cutoff like any other S-II…

indy91:
… engine failure; adjust S-II guidance timing presettings for that change


rcflyinghokie (approved):
Apollo 13 launch worked correctly with this

PR #1279, opened 2024-07-22
Overhaul of the RTCC lunar descent planning processor

indy91:
Testing to do: MCC calculations that use the LDPP


ThatRedMelon (approved):
All modes seem to give correct results

PR #1289, opened 2024-08-10
RTCC MFD: Improve VECPOINT display

indy91:
Works like Artemis universal tracking now and has optics/SIM bay pointing options.


ThatRedMelon (approved):
All options work as intended, tested on multiple bodies/spacecraft

PR #1294, opened 2024-09-06
MCC: Support for relay satellites and similar "ground" stations

indy91:


ThatRedMelon (approved):
Both relay satellites and ground stations provide a signal to the signal strength meter. The ATS satellite angles match the flight plan/EMP perfectly

PR #1298, opened 2024-09-23
Add new SimulatedTime variable in Saturn class, which makes it possible for MissionTime to be updated before launch

indy91:
Launch time can be changed in the PAMFD, via the GNC menu.

To accomplish this a new time parameter has been added, SimulatedTime. This time will always run continuously throughout the mission, without change. If the mission elapsed time is not changed before launch, essentially a launch delay, then SimulatedTime is identical to MissionTime. The SimulatedTime should be used for subsystems that need a steadily updating time through a whole mission, including saving/loading, but which does not need to have a specific time reference/start time.

MissionTime of the Saturn class can be changed before launch with this update, provided that the time is negative, so before liftoff. The majority of the changes of this PR make it so MissionTime is not used anymore by CSM subsystems and instead either simt from Orbiter or the new SimulatedTime parameter is used.

This PR is entirely backwards compatible as it initializes SimulatedTime with the current MissionTime, if SimulatedTime is zero.

EDIT: In order to support this capability for a Saturn IB based on docked stages, the following additional changes have been done:

-Simplify countdown logic. Only the mission time in the Saturn class (normal case) or the S-IVB (docked stages) determines when the launch happens. The extra logic in LC-34 and 37 for countdowns has been removed, including the LAUNCHMJD variable. In the future this should all be moved to the LCC class.
-S-IVB is now a supported vessel by the PAMFD and the launch time can be delayed there like for the Saturn class


indy91:
I'm working on some changes so that the liftoff time update works on LC-34 and LC-37 as well, it's currently broken there.

indy91:
I've made the additional changes for the Saturn IB consisting of docked stages and the PR is ready to review (again).

maxq519 (approved):
Works great

PR #1299, opened 2024-09-23
Add remaining modes of the RTCC Translunar Injection Processor

indy91:


ThatRedMelon (approved):
All modes work as described, giving what I understand to be correct solutions, including transfer to the MPT. Fully tested an Apollo 17 TLI-0, and ended up with a 3.2 fps MCC-2.

PR #1300, opened 2024-09-23
Add support for custom user-defined payloads in the Saturn 5/1B SLA, as well as an optional wide ELS-type SLA

maxq519:
Meshes created by @TINYBOB1


TINYBOB1:
image
image
The Wide SLA

maxq519:
This adds several new scenario options:
WIDESLA (0/1) - Use the wide ELS-type SLA
S4PL 9 - User defined custom payload
CUSTOMPAYLOADMASS (double) - Mass in kg of the custom payload
CUSTOMPAYLOADCLASS (string) - Specify the vessel config file to be loaded at payload creation

In order for a vessel to be compatible with this feature, it MUST have a docking port at the point where it attaches to the SLA.  This docking port MUST be the 2nd port defined for the vessel.

Note:  Wide SLA is not supported on the Saturn 1B vehicle, but custom payloads are.

ThatRedMelon (changes requested):
All scenario parameters work as defined, however, the sim crashes upon wide SLA separation, due to a missing vessel class. Are some config files missing in the PR?

maxq519:
Yep, I just missed them.  Should be fixed now

ThatRedMelon (dismissed):

Orbitersdk/samples/ProjectApollo/src_saturn/sivb.cpp
```diff
@@ -1736,6 +1815,10 @@ void SIVB::CreatePayload() {
 	case PAYLOAD_ASTP:
 		plName = "ProjectApollo/ASTP";
 		break;
+	case PAYLOAD_CUSTOM:
+		if (customPayloadClass == "") { strcpy(customPayloadClass, "ProjectApollo/ASTP"); } //Avoid weird stuff in case the user doesn't load a vessel class
```
indy91: customPayloadClass is a char array, can you even do == "" with it? And plName = customPayloadClass;?

maxq519: > And plName = customPayloadClass;?

This is the variable that it already uses to load the LM or ASTP DM vessel class.  (Line 1853)  Idk why

> customPayloadClass is a char array, can you even do == "" with it?

Looks like you can't, I'll fix that

indy91: I meant that " plName = customPayloadClass" has a similar issue and might need to be strcpy instead. Char arrays, always causing trouble, at least to my brain.

maxq519: It didn't work right when I tried to use strcpy
Just looked it up, it does appear to be valid.

indy91: Actually no, plName is just a pointer instead of a char array. Argh, I've never properly understood these string definitions :D

maxq519: Yeah they are really confusing and annoying

indy91: I now think plName = customPayloadClass is fine. plName will be pointing to customPayloadClass  and gets used later in the function, so it will give the right thing to oapiCreateVesselEx. I'll give it a test and then the PR should be good, the rest of the code looks fine.

maxq519: Sounds good

indy91 (approved):

PR #1301, opened 2024-10-03
Better handling of near parabolic orbits in Moon-centered Return-to-Earth conic subprocessor

indy91:
This is a change deep down in the moon-centered RTE logic, removing most code handling near parabolic orbits as a special case. This PR streamlines that logic a lot, only one place has to differentiate between the different orbit types now. As this change is so deep down on the RTE logic this PR could use a bunch of testing, any sort of Abort Scan Table calculations that are in the lunar sphere of influence.

Attached is a scenario 33 seconds before a TEI calculation that fails in the current NASSP build, but is fixed in this PR.

[A10 Pre TEI mcc bug 0002.scn.txt](https://github.com/user-attachments/files/17247734/A10.Pre.TEI.mcc.bug.0002.scn.txt)


ThatRedMelon (approved):
All RTE modes give the same solutions before and after the update, and the bugged scenario provided in the PR is fixed, giving a reasonable TLI-32 solution. (Both with and without MPT)

PR #1304, opened 2024-10-15
Add mission specific lunar marker files, change color to Cyan

rcflyinghokie:


ThatRedMelon (dismissed):
New colour is an improvement, and all landmarks appear where they should

ThatRedMelon (approved):
Inidividual markers work in both OpenOrbiter and Beta, in their correct locations. The main file is still the only one enabled by default. All markers appear in correct place and the colour is now cyan as intended, in my opinion better than the previous green.
image

PR #1307, opened 2024-10-16
Add working flashlight in the VC for both vehicles

maxq519:
"F" key toggles the flashlight on/off
Thanks to @Jordan-64 for helping with the code to get camera direction in local vessel frame!

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/saturnvc.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/saturnvc.cpp
@@ -805,6 +805,12 @@ bool Saturn::clbkLoadVC (int id)
 	vcFreeCamy = 0;
 	vcFreeCamz = 0;
 
+	//Flashlight
+	DelLightEmitter(flashlight);
+	flashlight = (::SpotLight*)AddSpotLight(flashlightPos, flashlightDirLocal, 3, 0, 0, 3, 0, RAD * 45, flashlightColor, flashlightColor, flashlightColor2);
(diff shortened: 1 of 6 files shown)


ThatRedMelon (approved):
The flashlight works beautifully in both R90 and OO. The brightness and size seems good in my opinion. The only issue is the single frame where the flashlight is in the incorrect position as discussed on discord, but if that can't be fixed, I'll approve it in the meantime, since everything else works and it is a really nice addition.

Orbitersdk/samples/ProjectApollo/src_csm/saturnvc.cpp
```diff
@@ -5022,3 +5028,31 @@ void Saturn::SetCMVCIntegralLight(UINT meshidx, DWORD *matList, int EmissionMode
...
+
+void Saturn::SetFlashlightOn(bool state)
+{
+	flashlight->Activate(state);
+	flashlightOn = state;
+}
+
+void Saturn::ToggleFlashlight()
+{
+	if ((oapiCockpitMode() == COCKPIT_VIRTUAL) && (oapiCameraMode() == CAM_COCKPIT)) {
```
ThymoNL, line 5055: Should the flashlight be forced off if we're not in the cockpit here?

maxq519: No need, it doesn't cost frames if we're not in the VC since Orbiter isn't rendering the light source

ThymoNL (approved):

PR #1310, opened 2024-10-25
Rework spacecraft attitude controls when using keyboard

CaptainSwag101:
As the title says, this overrides Orbiter's default attitude control handling to implement multiple different attitude/rotation controller deflection values, and changes the default deflection to be just shy of each vehicle's respective trigger point for direct thruster fire, but still commanding maximum proportional rate where applicable.

To get full deflection of the attitude controller and command direct/hardover RCS firing in the CSM/LM respectively, hold the left or right Alt key while commanding rotation. To move the ACA out of detent in the LM without commanding RCS thrust, hold Ctrl and Alt while commanding rotation.

This allows for more intuitive and nuanced control when using the keyboard, and significantly reduces unintentional thruster fire and wasted fuel, which would always occur previously because Orbiter's default behavior commands 100% stick deflection from keyboard inputs. Therefore, procedures which enable the direct RCS/hardover switches can now be followed safely by keyboard users without the side effects of reduced control fidelity and excess fuel usage, or turning off said switches to work around the issue.

Furthermore, this PR fixes an error that was discovered in the CSM RHC simulation: The direct RCS switches were implemented, but their state was not being checked by the appropriate RCS code, which was instead directly checking the input deflection value for an incorrect deflection threshold of 10.5° instead of 11°.

This PR primarily affects the keyboard control method. However, VESIM and Joystick controls will be affected by the change to the CSM RHC proportional rate scaling.

Some related info: the direct coil trigger point is 11° deflection out of a maximum 11.5° in the CSM, and 12° deflection out of a maximum 13° in the LM, per each vehicle's AOH Volume 1.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/scs.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/scs.cpp
@@ -836,7 +836,7 @@ void GDC::Timestep(double simdt) {
 			pitchrate = pitchBmag->GetRates().y*cos(Attitude.x) - yawBmag->GetRates().z*sin(Attitude.x);
 			if (A9K3)
 			{
-				//Assumption: secant amplifier gets saturated at +/-80° yaw
+				//Assumption: secant amplifier gets saturated at +/-80Â° yaw
 				pitchrate *= min(5.75877, max(-5.75877, 1.0 / cos(Attitude.z)));
 			}
 		}
@@ -850,7 +850,7 @@ void GDC::Timestep(double simdt) {
 			//GDC align
 			//Factor 3.33 because:
 			//ASCP output: 19.1V * sin(error)
-			//BMAG output: 0.1V per °/s (so 5.729 V per rad/s)
+			//BMAG output: 0.1V per Â°/s (so 5.729 V per rad/s)
 			//ASCP output basically needs to be convered to an attitude rate in rad/s, so the factor should be 19.1 divided by 5.729
 			//Source: Apollo SCS Detailed Training Program
 			if (E0_503PR)
@@ -1095,7 +1095,7 @@ void GDC::LoadState(FILEHANDLE scn){
 ASCP::ASCP(Sound &clicksound) : ClickSound(clicksound)
 
 {
-	// These are the nominal values for a 72° launch azimuth.
+	// These are the nominal values for a 72Â° launch azimuth.
 	output.x = 162;
 	output.y = 90;
 	output.z = 0;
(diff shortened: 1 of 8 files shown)


CaptainSwag101:
I can't make this change from mobile at the moment, but please temporarily consider this PR unfinished. I want to expand its scope to enable the out-of-detent signal for the LM and potentially CSM, to allow for the on-surface thruster checks without using fuel.

n7275:
> I can't make this change from mobile at the moment, but please temporarily consider this PR unfinished. I want to expand its scope to enable the out-of-detent signal for the LM and potentially CSM, to allow for the on-surface thruster checks without using fuel.

I marked it as draft for you.

CaptainSwag101:
This PR is now ready for review! I have significantly revised the description above to detail all the changes made. ~~In particular I'd really appreciate some help looking into the minimum value for the CSM rotation hand controller proportional rate transducers. In the LM, proportional rate goes from 1 degree of deflection to 11 degrees. In the CSM it goes from ??? degrees to 10 degrees. The AOH volume 1 makes mention of the rate command signal being routed "through the RC breakout logic" but the schematics it references don't make it clear whether or not the breakout switches, located at 1.5 degrees of deflection, are actually the trigger for proportional rate to begin.~~ (resolved)

ThymoNL:
How are we going to communicate this to new and existing users?
I think it's good to make an article on the wiki about this and some sort of announcement on the forum also wouldn't hurt.

CaptainSwag101:
I definitely plan on making a forum post about this. My hope, also, is that this change will be mostly not noticed by users if they don't happen to see it mentioned. The default behavior under most circumstances will remain unchanged; only in edge cases such as engine burns will the behavior differ, and even then the behavior ought to be more desirable than it was before. I think the "NASSP keyboard controls" forum post also ought to be updated with this new information.

Orbitersdk/samples/ProjectApollo/src_csm/scs.cpp
```diff
@@ -4106,9 +4106,9 @@ void ECA::TimeStep(double simdt) {
 		rate_err.y = cmd_rate.y - (rate_damp.y + pseudorate.y);
 		rate_err.z = cmd_rate.z - (rate_damp.z + pseudorate.z);
 		
-		// sprintf(oapiDebugString(),"SCS: RATE CMD r%.3f p%.3f y%.3f ERR r%.3f p%.3f y%.3f",
-		//	cmd_rate.x * DEG, cmd_rate.y * DEG, cmd_rate.z * DEG, 
-		//	rate_err.x * DEG, rate_err.y * DEG, rate_err.z * DEG);	
+		 sprintf(oapiDebugString(),"SCS: RATE CMD r%.3f p%.3f y%.3f ERR r%.3f p%.3f y%.3f",
```
ThatRedMelon: Should this still be uncommented?

ThatRedMelon (commented):
Overall the entire PR works with the correct keybinding and expected results. I can't test the vesim or joystick implementation as I don't have one to try, but the numpad controls work perfectly. Seems all good on my end, but I'll save on approving until someone can test with a joystick/vesim, as well as the debug line I left a comment on.

Orbitersdk/samples/ProjectApollo/src_csm/scs.cpp
CaptainSwag101: No it should not, nice catch!

ThatRedMelon (dismissed):

indy91 (changes requested):
Just the initialization of the rhc_keyboard_deflection variables as we talked about on Discord, then this PR is ready to go!

indy91 (approved):

PR #1312, opened 2024-11-08
Implement docking probe temp sensor for telemetry

rcflyinghokie:


ThatRedMelon (approved):
Docking probe temperature is successfully transferred to the telemetry client when the cbs are closed and stays within the voltage limits. When docked, the temperature of the probe remains relatively close to the current cabin temperature, alternating slightly. 

When undocked, a simulated period of 80 hours (maximum lunar orbit undocked time) gave these results for the probe facing the sun, the probe away from the sun, and the CSM in free drift:
image

This is not perfect to the real world conditions, but it provides the best balance between heating and cooling that I could find.

PR #1313, opened 2024-11-19
Frame count should run from 1-50 not 0-49

n7275:
Requires telemetry client rebuild

https://github.com/orbiternassp/telemetryClient/pull/9

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/csm_telecom.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/csm_telecom.cpp
@@ -1995,7 +1995,7 @@ PCM::PCM()
 	wsk_error = 0;
 	last_update = 0;
 	last_rx = 0;
-	frame_addr = 0;
+	frame_addr = 1;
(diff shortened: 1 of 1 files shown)


thewonderidiot:
This looks good to me! Will try it with the real PCM when I have a chance.

Orbitersdk/samples/ProjectApollo/src_csm/csm_telecom.cpp
```diff
@@ -4921,8 +4921,8 @@ void PCM::generate_stream_hbr(){
 	if(word_addr > 127){
 		word_addr = 0;
 		frame_addr++;
-		if(frame_addr > 49){
-			frame_addr = 0;
+		if(frame_addr > 50){
+			frame_addr = 1;
```
thewonderidiot, line 4925: It looks like the LBR function duplicates this code -- the documentation isn't clear but I would assume this also counts starting from 1. I'll try to capture some LBR data from the real PCM.

PR #1316, opened 2024-11-23
Fix normals for the BPC and docking probe

maxq519:
Also fixed a minor graphics issue where D3D9 was not loading the GC config for the apex cover after jettison


ThatRedMelon (approved):
The docking probe shows a huge improvement in graphical quality compared to how it was previously, see before and after pic below. Additionally, no noticeable graphical issues occurred elsewhere.

PR #1322, opened 2025-01-06
Make the flag fall over if it is to close to the LM as on Apollo 11

maxq519:
Distance < 8.25m: guaranteed to fall
Distance > 13.5m: guaranteed to stand
Between these it has a chance to fall based on distance

These are my best guess based on this [https://www.nasa.gov/history/alsj/ApolloFlags-Condition.html](url) and LROC quickmap
a11	8.2m	fell
a12	17.3m	standing
a14	9.5m?	probably fell?
a15	13.6m	standing
a16	16.5m	standing
a17	19.7m	standing


ThatRedMelon (dismissed):
Tested multiple liftoffs with the flag placed at increasingly further distances from the LM. The flag falls correctly outside the safe/fall range, and randomly between them. The flag also saves/loads its state correctly.

rcflyinghokie (dismissed):

indy91 (changes requested):
This PR consistently uses bitwise AND `&` instead of the normal AND operator `&&` While I believe this technically gives correct behavior in C++, I still think it should be corrected.

The rest of the PR looks ok. I see some potential problems in the future in the code that avoids detecting the descent stage by accident, because it assumes the max propellant mass is exactly 0.2 kg. If anything is changed in the descent stage code then this will break.

maxq519:
> I see some potential problems in the future in the code that avoids detecting the descent stage by accident, because it assumes the max propellant mass is exactly 0.2 kg. If anything is changed in the descent stage code then this will break.

Yeah, I agree but I couldn't think of any other easy way to do this.  Even if this breaks in the future though, it is not a critical issue, it would only matter in a very specific edge case where the LM is staged without launching, then the scenario is exited and reloaded.

indy91 (approved):

PR #1323, opened 2025-01-06
Final Apollo 10 checklists to pair with MCC updates

rcflyinghokie:


rcflyinghokie:
not sure why these are conflicting, they can overwrite the current 10 checklists

PR #1324, opened 2025-01-08
RTCC: Changes to deorbit targeting to allow the two maneuver Skylab deorbit sequence to be calculated

indy91:
-Time between separation maneuver and deorbit maneuver can be any length of time (previous limitation 81 minutes)
-Deorbit targeting can calculate desired REFSMMAT
-Cleanup of deorbit targeting code
-Shaping/separation maneuver display shows perigee height and External DV vector
-Saving/loading of retrofire save table in scenarios (data for MEDs R30, R31 and R32 are stored there)
-Small fix to RTCC CSM/LM maneuver simulation for fixed DV maneuvers
-Add explanation of Skylab sequence to RTCC MFD manual


ThatRedMelon (approved):
Tested using SL-2 scenario from pre-launch to landing, shape/deorbit burn calculations were successful and gave correct values (see image below)

image

After shape was completed, de-orbit alone was calculated in the MFD without a separation maneuver included, and the calculations agreed with the pre-shape calculations (de-orbit was off by 0.05s, expected with shape residuals). Then entry was performed according to the checklist with no anomalies, for extra verification here is the entry PAD: 

image

PR #1325, opened 2025-01-12
Move more MCC only functions from the RTCC class to the MCC_Calculations class

indy91:
This PR only moves several functions, that are used by the MCC but are not real RTCC functions, from the RTCC class to the new MCC Calculations class. As a start of a more substantial restructuring it also does a small part of the mission initialization in a utility function now. This part I have tested.

This PR will cause some merge conflicts with work-in-progress MCC updates like #1321 but they should be easy to fix.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_launch/MCC_Calculations.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_launch/MCC_Calculations.cpp
@@ -119,3 +119,286 @@ bool MCC_Calculations::LongitudeCrossing(EphemerisDataTable2 &ephem, double lng,
 
 	return (dErr < 0.0);
 }
+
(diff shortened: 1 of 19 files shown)


rcflyinghokie:
> This PR only moves several functions, that are used by the MCC but are not real RTCC functions, from the RTCC class to the new MCC Calculations class. As a start of a more substantial restructuring it also does a small part of the mission initialization in a utility function now. This part I have tested.
> 
> This PR will cause some merge conflicts with work-in-progress MCC updates like #1321 but they should be easy to fix.

Merge conflicts should be resolved with the MCC updates branch

rcflyinghokie (approved):
Will resolve future MCC branch merge conflicts after this gets merged

PR #1328, opened 2025-01-14
Three RTCC changes

indy91:
-RTCC MFD: Converting any TLMCC or LOI maneuver to a finite burn (taking thruster into account) now uses the same inputs and RTCC functions as the MPT equivalent. This includes ullage selection, iteration to find the optimum TIG etc. This same change has already been done to many other RTCC processors. Testing would involve some TLMCC maneuver or LOI maneuver calculations.

-The RTCC has two (closely related) coast numerical integrators. The gravity model constants for these were now moved to the RTCC System Parameters file, so that they only exist in one place. Only two lines of code have to be changed now to switch from the gravity model we use for the Orbiter Beta as compared to the Open Orbiter gravity. This gravity model "order" is also loadable by the RTCC mission constants file (see RTCC MFD manual).

-In testing the gravity model change I discovered a small bug in the FDO Orbit Digitals calculation with the U12 MED (predicted apoapsis/periapsis). The only issue was that the wrong times for periapsis/apoapsis were displayed. This is now fixed.


maxq519 (approved):

ThatRedMelon (approved):
All affected functions perform as expected

PR #1331, opened 2025-01-19
Correct Sign Error in PIPA Scale

n7275:


indy91:
I was going to suggest deleting any "PSCALE" lines from the IMU section of mission scenarios so that they are loaded again from the mission file, but it seems like we never added them to the mission scenarios. So I think all that has to be done for this fix is to inform people that they have to do these deletions in their scenarios for any ongoing mission.

indy91 (approved):

PR #1334, opened 2025-01-22
Ability to calculate multiple DKI plans at once and more RTCC MFD Cleanup

indy91:
-Ability to calculate multiple DKI plans at once
-SPQ uses same calculation as DKI for transfer to finite burn
-Same button label that is used for this transfer on every page
-Most ullage selection buttons now use the same code
-Also a function for a uniform display of these ullage options
-Tweaks to default ullage duration selection


ThatRedMelon (approved):
All additions work as described both with/without MPT

PR #1335, opened 2025-01-23
LM Descent Stage J Mission Switching

TINYBOB1:
This adds the feature to switch between G/H and J mission decent stages using the "JMission" flag in the mission file.


TINYBOB1:
After much deliberation with @maxq519, we cannot find a fix for the deflector and cask flags causing a crash at stage sep.  Max believes that this is due to how the descent stage vessel is coded, loading the mesh before it loads the mission file to see if the mission is a J one, and thus cant reload the visual to use the new mesh at staging.  using this for the IRL as flown missions, it works fine, since no J class LM ever had removed deflectors or a removed cask.  I can finish this PR as is, but wanted yalls opinion on this.

PR #1339, opened 2025-01-28
Fix simulated S-II engine failures that are set before S-II engine st…

indy91:
…art; save S-II engine failure condition; various scheduled engine failures can be reset before they occur

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_saturn/j2engine.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_saturn/j2engine.cpp
@@ -168,10 +168,12 @@ void J2Engine::Timestep(double simdt)
 		ThrustTimer = 0.0;
 	}
 
+	/*
 	if (EngineFailed)
 	{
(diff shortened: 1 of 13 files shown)


Orbitersdk/samples/ProjectApollo/src_saturn/j2engine.cpp
```diff
@@ -146,23 +147,31 @@ void J2Engine::Timestep(double simdt)
...
+		vessel->SetThrusterLevel(th_j2, ThrustLevel);
 	}
 	else if (ThrustTimer > 0.0)
 	{
 		ThrustTimer = 0.0;
 	}
+
+	if (EngineFailed)
+	{
+		sprintf(oapiDebugString(), "EngineStart %d EngineStop %d EngineRunning %d ThrustTimer %lf ThrustLevel %lf", EngineStart, EngineStop, EngineRunning, ThrustTimer, ThrustLevel);
```
ThatRedMelon, line 174: Should the debug string be left in?

indy91: ouch

ThatRedMelon (approved):
All stages fail/save their failed state correctly

PR #1344, opened 2025-02-02
Improved Models for Substance Bulk Modulus/Liquid Compressability

n7275:
Also, fix all the whitespace in Hsystems.cpp, because I committed without checking after pressing some VS key combination...


rcflyinghokie:
You left debug lines in place just fysa :)

rcflyinghokie:
Other than the debug lines still there, this was very stable on Apollo 8. I have numerous saves if you need data points for review.

rcflyinghokie (dismissed):
Have tested this on 3 flights, should be good to go.

rcflyinghokie (approved):

PR #1353, opened 2025-02-18
Fix rare cases of TEI calculation not converging; making a few variab…

indy91:
…les safer and removing a bit of unneeded code in the moon-centered RTE logic


ThatRedMelon (approved):
In all cases I've tested, TEI has calculated correctly, including where it didn't previously.

PR #1355, opened 2025-02-20
Project Apollo MFD: GNC page shows parameters only relative to Earth …

indy91:
…and Moon, not the Sun, using the same lunar SOI radius as the AGC


ThatRedMelon (approved):
Correctly displays all values in both reference bodies, as well as transitions between the two without going to Sun reference.

PR #1356, opened 2025-02-20
Correct the Skylab empty mass value (it had included the payload shro…

indy91:
…ud and TACS propellant)


rcflyinghokie (approved):
Math is correct :)

n7275 (approved):

PR #1358, opened 2025-02-21
Add all Skylab launch targeting inputs to the RTCC MFD inputs page; s…

indy91:
…everal internal LWP corrections


ThatRedMelon (approved):
I believe all simulated inputs work as they should (to my understanding), giving sensible results. Transferring to the MPT works correctly and uplinking succeeds too.

PR #1362, opened 2025-02-23
CMVC Fix Emission Texture Line not matching with main texture in Mission Timer.

Jordan-64:


Jordan-64:
This is the bug reported by @CaptainSwag101 
grafik

indy91 (approved):

PR #1365, opened 2025-02-27
Limit EMS DV counter in all cases

ThatRedMelon:
Previously the EMS DV counter had a limit on the minimum and maximum values, but only in DV or backup modes. This meant that you could use the DV-Set function (and maybe others) to go out of the limits. This PR changes the EMS to enforce this limit in all circumstances.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/scs.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/scs.cpp
@@ -4932,7 +4932,6 @@ void EMS::TimeStep(double simdt) {
 			// dV/Range display
 			if (xacc > 0 || status == EMS_STATUS_DV) {
 				dVRangeCounter -= dV;
-				dVRangeCounter = max(-1000.0, min(14000.0, dVRangeCounter));
 			}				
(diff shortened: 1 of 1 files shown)


rcflyinghokie:
Shouldn't this also allow @CaptainSwag101 to change the EMS display to the new EL digits?

rcflyinghokie (approved):

CaptainSwag101:
> Shouldn't this also allow @CaptainSwag101 to change the EMS display to the new EL digits?

This wasn't anything that was blocking my ability to use the new digits, in fact I didn't even know about this bug. The digits just don't fit the display properly. I can give it one more shot but we're so limited on space that I don't think it'll work very well.

indy91 (approved):

rcflyinghokie:
> > Shouldn't this also allow @CaptainSwag101 to change the EMS display to the new EL digits?
> 
> This wasn't anything that was blocking my ability to use the new digits, in fact I didn't even know about this bug. The digits just don't fit the display properly. I can give it one more shot but we're so limited on space that I don't think it'll work very well.

I thought there was enough space because the leading digit was only a 1 or a -

Issue #1319, opened 2025-01-04
CSM DC Voltameter and Ammeter issues


franzfrancinehopper:
- [ ] 1) Batt C should be able to be disconnected from volt meter after pulling Bat C PWR Entry/ Post Landing (CB 22) or Bat C Bat CHGR / EDS 2 (CB 23)

- [x] 2) Fuel Cells should not have volt readout when selected on DC Indicators, there is no connection.

- [x] 3) Pyro Bats A and B should not get an amp readout when selected on DC indicators.

- [x] 4) Main A and Main B should not have an AMP readout when selected with DC indicators

- [ ] 5) Disconnecting any EPS Sensor Signal CBs does not disconnect the Voltmeter for any corresponding bus.

- [ ] 6) Battery relay bus volts need to be read/routed through the FC BUS CONT 1 cb

- [ ] 7) Pulling BAT CHG A or B cbs should disconnect reading to the BAT BUS VOLTS sensor and thus zero the voltmeter  

- [ ] 8) Entry batteries should have an "inaccessible" 100A cb between the battery and its first circuit breaker connection.

rcflyinghokie:
2,3,4 fixed, 1 and 2 need more extensive power handling in V9

rcflyinghokie:
Additionally the 3 entry batteries should have an "inaccessible" 100A cb between the battery and its first circuit breaker connection.

PR #1366, opened 2025-02-28
Correct CSM Volt & Amp Meter Wiring

rcflyinghokie:
#1319 Issues 2, 3, 4 fixed so far
References:
image
image
image


ThatRedMelon (approved):
If this PR only aims to fix issues 2,3, and 4 of #1319, then it is successful in doing so, and no other issues were experienced

rcflyinghokie:
> If this PR only aims to fix issues 2,3, and 4 of #1319, then it is successful in doing so, and no other issues were experienced

Yep fixes those for now, the others will take a more extensive rewiring since they involve two way power, the git issue will remain open indicating those remaining

indy91:
Part 1 of issue #1319 will require a more complicated rewiring, so I will merge this PR as it is, but the issue is not fully solved.

PR #1367, opened 2025-03-01
Saturn V LVDC can use the S-II for orbit insertion (Skylab); slightly more precise TLI cutoff; a few minor LVDC corrections

indy91:
This update makes it possible to have a guided orbit insertion by the Saturn V LVDC using the S-II. The IGM stages are different in the following way:

Normal Saturn V:
-IGM stage 1: S-II before mixture ratio shift
-IGM stage 2: S-II after mixture ratio shift
-IGM stage 3: S-IVB to orbit insertion

Skylab:
-IGM stage 1: S-II before center engine cutoff
-IGM stage 2: S-II before mixture ratio shift
-IGM stage 3: S-II after mixture ratio shift

At S-II cutoff the LVDC goes from timebase 3 through timebase 4 and like the normal Saturn V sequence uses timebase 5 for Earth orbit. Several presettings for the IGM have to be changed to allow this guidance.

Also in this update is a small change in the flight sequence programs, in the switch selector sequence for the normal S-IVB cutoff (Earth orbit insertion). It used a wrong return from the alternate sequence and started timebase 5 in a wrong way. Also several small interrupt processing corrections are included in this update.

For testing the S-II to orbit insertion I have used a modified Apollo 8 scenario and the Skylab flight sequence program (so far without any SWS commands):

[Apollo 8 - 02 - Before Launch T-60s - S-II Insertion Test.scn.txt](https://github.com/user-attachments/files/19039090/Apollo.8.-.02.-.Before.Launch.T-60s.-.S-II.Insertion.Test.scn.txt)
[AS-513 Flight Sequence Program.txt](https://github.com/user-attachments/files/19039091/AS-513.Flight.Sequence.Program.txt)

EDIT: As part of a bugfix I am also including a small new feature for TLI. The correct S-IVB cutoff sequence is now used, which very slightly makes cutoff precision better. Instead of commanding the cutoff at 0.04 seconds with an average delay of actually issuing the command about 0.04 seconds after that the cutoff is only issued 0.02 seconds "too early".


indy91:
I did another launch through TLI with this PR, should be good to test now. Aside from this testing scenario the S-II to cutoff feature will be rather difficult to test with any other scenario, so mostly testing for this PR should consist of making sure that the Saturn V LVDC still does TLI properly and just an overall check on the Saturn V LVDC not behaving differently in any way.

ThatRedMelon (approved):
S-II orbit insertion works correctly, and there are no added issues with regular flights.

PR #1371, opened 2025-03-03
Fix MFD related crashes caused by checklists not fully loading

lassombra:
This is not the cleanest approach but is the most surgical change I could find to fix this bug.  It ensures that even if no checklists are loaded, the checklist MFD doesn't loop until `TopStep` overflows and goes negative causing an access violation since `TopStep` is an index.


rcflyinghokie:
While this might make it harder to see immediately that a checklist is an issue, it will prevent CTD for releases.

lassombra:
I think I'm going to put this to draft and add two things:

1) A fix for crashing when a checklist is in progress
2) a configuration option to throw an exception if checklists fail to load

This will improve behavior for users, but give Folger's the quick crash behavior.

PR #1375, opened 2025-03-05
Remove Obsolete/Depreciated/Unnecessary Documents, Add Quick Start Guide

rcflyinghokie:
Remove files no longer used or no longer applicable to be shipped with NASSP to reduce size and confusion. Add quick start guide for reference with links.


ThatRedMelon (commented):
I am happy for all of these to be deleted, and I agree that the A7 Update Forms should stay. I won't give explcit approval to let everyone else agree on this too

indy91 (approved):

PR #1376, opened 2025-03-05
RTCC MFD: Simplify REFSMMAT uplink options (no distinction between MPT and non-MPT mode) and make all RTCC REFSMMATs available for uplink

indy91:
Previous behavior:
-In MPT mode the C12 MED has to be entered in correct format. All REFSMMATs available and actual vs. desired can be selected
-In non-MPT mode only actual vs. desired can be selected by entering 1 or 2. Only the CUR REFSMMAT is selected.

New behavior:
-In both modes all REFSMMATs are available for uplink. The MFD input box is prefilled with "CUR 2" to uplink the current REFSMMAT as the desired one, but these inputs can be changed.


ThatRedMelon (approved):
Works as intended, and a nice addition!

PR #1383, opened 2025-03-09
Add cryo tank pressure relief valves based on CSM 104 values

rcflyinghokie:


ThatRedMelon (approved):
Pressure relieves as expected

PR #1385, opened 2025-03-12
Fix yaw keyboard inversion in LM

CaptainSwag101:
This fixes an issue introduced in #1310 where the LM yaw controls using keyboard were inverted from how they were before.


ThatRedMelon (approved):
Works as intended

PR #1386, opened 2025-03-15
Fix two RTCC bugs

indy91:
-Splashdown coordinates downlink from the CMC to the RTCC on the Entry PAD page was broken
-Present position data calculation for FDO Orbit Digitals calculates orbital period even for a hyperbolic orbit, which could lead to a graphics CTD when trying to show this number


ThatRedMelon (approved):
All good!

PR #1387, opened 2025-03-17
Change Apollo earth and add moon markers to use label format 2, change resolution from 04 to 01

rcflyinghokie:
- Changes the earth markers so they will display in TLC
- Adds label format 2 moon markers for Apollo 8-12
- Color on O16 beta and O24 is cyan due to an RGB inversion that was corrected in the latest OO. Color will be yellow for the latest OO release with this fix. Color can be changed once a stable O24+ release for NASSP is completed.


ThatRedMelon (approved):
Markers appear in their intended position, aside from an Orbiter bug that isn't a fault of NASSP.

PR #1388, opened 2025-03-20
Floatation Collar

TINYBOB1:
Adds the flotation collar that would be installed by the recovery team after splashdown.
image


ThatRedMelon (approved):
Looks great!

Issue #1117, opened 2023-12-01
Lat/Lon within Apollo Navigation Landmarks.mkr should be updated


Aeasala:
When visual markers have been enabled within orbiter, the lat/lon of landmarks defined in this file do not match up with the actual landing sites showcased in documents.

For example, CP1 thru CP3 of Apollo 8 are many, many miles off.

Correcting the table to the lat/long shown in **Table V** (p. 21) of _Lunar Landmark Locations – Apollo 8, 10, 11, and 12 missions_ resolved their locations.

It is believed that the current values may have been derived from **Table IV** of the mentioned document, which are position 'solutions' for the tracking sequence.

Aeasala:
I might be mistaken about the last note.
Either way, example of current selected values:
-162.700 -5.250        : CP1 504 (Apollo 8)
+155.100 -10.200     : CP2 526 (Apollo 8)
+95.9 -9.1              	: CP3 334 (Apollo 8)

From Table V:
-158.0462 -6.3079        : CP1 504 (Apollo 8)
+163.2463 -9.7036     : CP2 526 (Apollo 8)
+96.8915 -8.8990   	: CP3 334 (Apollo 8)

indy91:
I will go through the coordinates some time and make necessary corrections. Thanks!

indy91:
It's interesting. The coordinates we have been using are from the flight plan, so theoretically they should be good. We don't have the full Apollo 8 Lunar Landmark Maps document available but I found one page and the coordinates differ from the flight plan:

Flight Plan:

![image](https://github.com/orbiternassp/NASSP/assets/13571607/e52f2287-fad6-4481-a587-b370071a3ee6)

Landmark Maps:

![image](https://github.com/orbiternassp/NASSP/assets/13571607/f857672c-8acf-4919-a717-531e09d21247)

So our values should definitely be changed. Maybe not necessarily to the ones that got used on the actual mission, as the exact coordinates were more of a suggestion than exact. But we should definitely use the coordinates that Jim Lovell intended to use before the mission. I'll do some more research!

Aeasala:
Yeah, not sure about how it should go forward, but the document I listed does provide all of the landmark coordinates to a T, geographically.  At least when comparing the visual references provided in the document for each landmark, there's pretty much zero error.

I'm assuming this .mkr file is only for visual aid if markers are enabled, and not actually referenced behind the scenes as NASSP does its thing.  If that's not the case, I like the idea in your last note.

[...]

PR #1389, opened 2025-03-20
Make lunar landmark markers more accurate 

ThatRedMelon:
This PR changes the location of all of the lunar landmark markers to line up with the actual position and elevation from the lunar landmark maps of each mission, as well as adding all landmarks from Apollo 13 to Apollo 17. The only exceptions from using the exact position is landmark H-1 from Apollo 12, and all landmarks from Apollo 16 (excluding F-1, since that is reused from Apollo 10), since these landmarks don't have an available map to cross reference. I'd appreciate some tests on both the regular and mission specific markers to make sure nothing is obviously out of place or floating/sunken in the ground.

This PR will solve #1117


rcflyinghokie:
These look great!

rcflyinghokie (approved):

PR #1393, opened 2025-03-27
MCC: Reduce text size of annotations from 0.65 to 0.60 scale

indy91:
This allows additional lines of remarks to be shown for Maneuver PADs and prevents some cases of overlap between the CAPCOM menu and the PADs.

Before:
image

After:
image


rcflyinghokie (approved):
Works great and leaves extra room for remarks which has proven to be useful

ThatRedMelon (approved):

maxq519 (approved):

PR #1398, opened 2025-03-31
Backup GDC alignment changes

indy91:
-Rigel, Sirius available as a 4th star pair
-In the CSM Maneuver PAD options a preferred star pair for the backup GDC alignment can be chosen using the variable PrefGDCStars
-Sextant star check defaults to 20 minutes before TIG (MCC only)
-Backup GDC alignment is calculated using the sextant star check time


rcflyinghokie (approved):
Works as intended!

PR #1399, opened 2025-04-01
Apollo 7 MCC: Fix NCC-2 calculation; initialize two impulse processor results to zero

indy91:
Forgot to convert input times from GET to GMT. Attached is the scenario that has a broken Maneuver PAD and gives bad/random results when trying to repeat the NCC-2 calculation.

[Apollo 7 NCC-2 Broken PAD.scn.txt](https://github.com/user-attachments/files/19553865/Apollo.7.NCC-2.Broken.PAD.scn.txt)


ThatRedMelon (approved):
Works as intended

PR #1406, opened 2025-04-07
RTCC: Cleanup of PDI simulation related code

indy91:
Purpose of this PR is mostly to remove using MJD in calculations as well as the old state vector type (SV).

Things to test would be:

-PDI PAD
-PDI on MPT
-Powered Descent Abort Program
-LDPP calculation of PDI time

Especially GMT/GET conversion bugs could come from this change.


ThatRedMelon (approved):
Works as intended!

PR #1407, opened 2025-04-08
Add NASSP-specific quicksave feature with more useful save names

maxq519:
Name format:  Apollo 11 - 69h 0m 5.49s
New keybind:  Ctrl+Alt+S

This does _not_ affect Orbiter's default Ctrl+S quicksave functionality in any way.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
@@ -2003,6 +2003,36 @@ void Saturn::clbkSaveState(FILEHANDLE scn)
 	eventControl.save(scn);
 }
 
+void Saturn::QuicksaveScenario()
+{
+	double time = MissionTime;
+	VECTOR3 hhhmmss = _V(0, 0, 0);
+
+	for (time; time > 3600;) {
+		time -= 3600;
+		hhhmmss.x += 1;
+	}
+	for (time; time > 60;) {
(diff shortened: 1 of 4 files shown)


n7275:
Any way you can force pausing of the simulation before calling the save function. Maybe it's just because I have slow disks, but I often get a really long timestep when saving.

indy91:
> Any way you can force pausing of the simulation before calling the save function. Maybe it's just because I have slow disks, but I often get a really long timestep when saving.

You might have to clean up your saves folder and back it up. Orbiter does some badly performing things when saving, I think it is l looking through all files in the saves folder (maybe only quicksave folder? Not sure). But this is the main cause of a slow saving.

maxq519:
> Any way you can force pausing of the simulation before calling the save function. Maybe it's just because I have slow disks, but I often get a really long timestep when saving.

Pausing the sim would be easy, un-pausing it not so much.  Since all the save code runs in a single timestep Orbiter can't process both a pause event and an unpause event.

n7275:
> > Any way you can force pausing of the simulation before calling the save function. Maybe it's just because I have slow disks, but I often get a really long timestep when saving.
> 
> Pausing the sim would be easy, un-pausing it not so much. Since all the save code runs in a single timestep Orbiter can't process both a pause event and an unpause event.

This may be something to add into the Orbiter core then. I'll look at feasibility.

indy91 (approved):

PR #1412, opened 2025-04-12
Skylab 2 Updates

GoForPDI:
Add checklists through docking to Skylab 2, and add one missing state vector update to its MCC. Checklists are based off the SL-3 launch checklist and SL-4 rendezvous book.


maxq519 (changes requested):
Looks like lots of files are misplaced, they are showing up as new files

GoForPDI:
git screwed up it seems - working on fixing it with indy rn

PR #1415, opened 2025-04-12
Add XRSound config for LRV

MrResiduals:
I added an XRsound config for the LRV since I noticed it didn't have one. I used the LEVA config as my base.


maxq519 (changes requested):
Looks like I still hear music and radio sounds in external view

GoForPDI:
this makes sense given using the LEVA config as a base - that also has music/radio in external

maxq519 (approved):
Looks like I had the wrong branch, everything works great

indy91 (approved):

PR #1416, opened 2025-04-13
Three small RTCC fixes

indy91:
-Space Digitals column 1: Show vehicle area and K-factor and fix eccentricity being shown instead of inclination
-Columns 5 and 6 of the midcourse tradeoff can now be deleted
-When PDI is moved to the MPT and the ignition algorithm fails then shown an error message and stop processing


ThatRedMelon (approved):
All fixes work!

PR #1420, opened 2025-04-17
Fix S-IVB venting in Apollo 12-14 flight sequence programs

indy91:
This PR fixes a few issues with S-IVB venting on Apollo 12-14.

On Apollo 12 the LH2 non-propulsive venting valves were not opened. Probably causing a bit too much of the H2 to be converted into an undesired thrust.

More importantly, in the flight sequence program file that we use for Apollo 13 and 14 only the LH2 continuous vent system (CVS) is opened in Timebase 8, not the LH2 non-propulsive vent. This causes a thrust of about 1-2 Newton to affect the trajectory of the S-IVB nearly all through translunar coast, even leading to the S-IVB missing the Moon instead of hitting it. The changes in this PR open and close all the LOX and LH2 vents at the correct times in TB8.

I am also starting to add some of these switch selector sequences to our wiki: https://nassp.space/index.php/Flight_Sequence_Program

Note for testing: Start before TB8 because old scenarios in TB8 might have a relay in the wrong state for opening valves.


ThatRedMelon (approved):
Venting works as intended, with the vents closing at the right point after TB8

PR #1421, opened 2025-04-17
Add tracking lights to Skylab and floodlight vessel to the launch scenario

maxq519:


ThatRedMelon (approved):
Tracking lights work as expected from a distance of 400nm or less.
Flodlights work as expected on launch site

rcflyinghokie (approved):

PR #1422, opened 2025-04-18
Two RTCC changes

indy91:
-Lunar impact targeting now has three modes. Trajectory evaluation (current trajectory, no burn), burn evaluation (impact conditions of input burn) and the normal burn calculation. Burn optimization can be turned on and off. Also the impact flight-path angle is displayed.
-Detailed Maneuver Table had a sign error in the displayed DH and phase angle (theta). Also too complicated code for their calculation. Some greek letters are now used on the display and the FDAI attitude is blanked if the maneuver is not a LM maneuver.
-Both of these are now properly documented in the RTCC MFD manual.


ThatRedMelon (approved):
Lunar targeting gives correct calculations and works with/without optimization. DMT gives correct phase and DH sign, and everything else works without change

Issue #996, opened 2023-05-24
LM LCA Implementation 


rcflyinghokie:
Almost all lighting in the LM gets controlled through the Lighting Control Assembly (LCA) which routes DC and AC power from ANUN/DOCK/COMP, INTEGRAL, and NUMERIC lighting breakers to the rheostats and override switches which control various lighting around the LM. This device is essentially a power handler and generates heat and is attached to a cold plate (implemented.) This should be implemented to handle the power routing of lighting in the LM instead of wiring directly to breakers.

This also will be used to change power control for the LM DSKY #972 and add a missing "heater" for #233 .

rcflyinghokie:
#1628

PR #1424, opened 2025-04-20
LMVC side panels  integral lights can be switched off independently of the others.

Jordan-64:
The integral lights of the side panels in the LM VC can be switched on and off independently of the others, using the respective switches in panels 5 and 3.


ThatRedMelon (dismissed):
If my understanding of the side panel lights is correct, this PR works as it should

rcflyinghokie:
This PR doesnt seem to look for power, only the switch position. Needs power connections and appropriate draws added

Jordan-64:
That's right, once you've finished your code, we can update it. I made the PR so we have something that works, otherwise you wouldn't be able to test your code.

rcflyinghokie:
Works well enough visually, we can add LCA power considerations later #996

rcflyinghokie (approved):

PR #1425, opened 2025-04-20
RTCC improvements concerning the stored AGS REFSMMAT

indy91:
-Fix bug in saving the OST REFSMMAT using the LOST display in AGS mode, using the backup AGS alignment technique with the AOT
-AGS State Vector PAD page now has the option to use any REFSMMAT for the state vector calculation
-AGS State Vector PAD page additionally shows the AGS REFSMMAT permanently because on the vector panel summary when an AGS state vector is saved from telemetry only the AGS REFSMMAT stored in the RTCC is used. There is now also a button to save the LM CUR as the LM AGS REFSMMAT, as a convenient way to get this AGS REFSMMAT. Previously the G00 MED to move the REFSMMAT would have to be used.
-When saving an AGS state vector in the vector evaluation table the conversion from AGS to inertial coordinates (via the AGS REFSMMAT) is now done when the the vector is moved to the evaluation table. Previously the conversion was done when the telemetry vector was saved (TLM button).
-If the AGS REFSMMAT is not available when trying to move the AGS state vector to the evaluation table, print an error message on the on-line monitor
-Update RTCC MFD manual with all these updates


lassombra:
Does all the things it reports to do - no obvious issues

rcflyinghokie:
not sure why I can add lassombra as a reviewer, will approve based on testimony :)

rcflyinghokie (approved):

PR #1427, opened 2025-04-22
Make Vector Panel Summary vector movements safer; and two other small RTCC fixes

indy91:
Adds several error codes and error returns to VPS vector movements:
-Telemetry to Evaluation Vector Table if telemetry vector is not available
-Telemetry to Evaluation Vector Table if telemetry vector is below radius of body (Earth or Moon)
-Evaluation Vector Table to Usable Vector Table of vector is not available
-Usable Vector Table to Ephemeris Update if vector is not available
-Additionally any VPS vector movement is run in a thread, so that e.g. a failing trajectory upate does not lock up Orbiter

Fixes two other small issues:
-If the selected LM vessel in the RTCC MFD is not actually a LM do not try to get the AGS telemetry state vector from it
-Apollo 11 MCC: Use a different method to get the TLI+90 TIG from the TLI ignition time. The new TIG is much closer to the TLI+90 on the actual mission than before

### DIFF
--- a/Doc/Project Apollo - NASSP/RTCC/ApolloRTCCMFD.pdf
+++ b/Doc/Project Apollo - NASSP/RTCC/ApolloRTCCMFD.pdf
@@ -823,92 +823,96 @@ endobj
 (Display Parameters and Buttons)
 endobj
 549 0 obj
-<< /S /GoTo /D (section.14) >>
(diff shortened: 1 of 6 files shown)


lassombra (approved):
Works as advertised - couldn't test below surface state vectors but all other scenarios mentioned are working as reported.  Errors appear correctly on online monitor

PR #1432, opened 2025-04-23
RTCC MFD: Move sextant star check time option to Maneuver PAD page; M…

indy91:
…aneuver PAD has option to select preferred star set for the backup GDC alignment


ThatRedMelon (approved):
All set stars give correct angles and chooses a separate set when occulted/has too large an MGA. Star check time gives correct calculations depending on the time.

PR #1434, opened 2025-04-25
A few FDO Orbit Digitals fixes and default Vector Panel Summary DV vehicles

indy91:
This PR contains:

-Fix FDO Orbit Digitals calculation of present apogee/perigee data
-Also fix longitude at a time (U14) calculation for Earth orbit
-Add some error codes for the FDO Orbit Digitals
-RTCC MFD: Vector Panel Summary uses selected CSM/LM vessels on config page by default for the DC solution


ThatRedMelon (approved):
All features work without issue

PR #1435, opened 2025-04-26
Fix flood lights not working correctly when the VC is loaded while being paused; remove function updating the flood light position on every timestep as it is not required

indy91:
Removing the UpdateFloodLights functions should be fine as the flood light positions are automatically updated by Orbiter with any CG shift. This could use some testing though, to confirm it works correctly in all cases. Test cases would be:

-Saturn launch
-Any staging
-Powered descent
-Long SPS burn


ThatRedMelon (approved):
Floodlights behaved as expected without fail during:

| Items Tested |
| --- |
| Loading While Sim Paused |
| Switching Views While Sim Paused |
| Full Saturn Launch |
| CSM/LV Sep |
| CSM/LM Sep |
| CSM/SM Sep |
| Ascent/Descent Stage Sep |
| LOI Burn |
| PDI Burn |

PR #1437, opened 2025-04-28
A few RTCC changes

indy91:
-RTCC MFD: Show MOCR displays centered in the MFD screen
-Fix right ascension input in G13 MED and output on GOST, LOST and AGOP (hours instead of degrees)
-Add RTCC star catalog numbers to marker file with additional RTCC stars

A fair amount of changes were required to the RTCC MFD display code, so all MOCR displays in the MFD should in theory be tested

EDIT:

Two more minor fixes have been added to this PR:
-Actually use selected REFSMMAT for AGS SV PAD calculation
-On two impulse processor transfer page the selected table (multiple solution vs. corrective combination) can be chosen


ThatRedMelon (approved):
All new displays and features work as intended with no obvious issues

PR #1438, opened 2025-05-02
RTCC thrust tables

indy91:
This PR adds the thrust tables that are used in the main thrust phases for SPS, DPS and APS maneuver simulation. The tables are used in two places, the general CSM/LM maneuver simulation routine (for e.g. External DV burns) and also the lunar ascent simulation. The PR also adds the two MEDs (M10 and M11) that change the thrust and CG tables. Descriptions of the MED input were added to the RTCC MFD manual. The tables can be accessed via the configuration page of the MFD.

For testing purposes, this change affects the following:

-Iterate option on a maneuver transfer performs a full maneuver simulation, so the result will be slightly different depending on the thrust tables
-Maneuver PAD for burns with SPS, APS, DPS
-Ascent maneuver on the mission plan table
-Lunar ascent processor
-Powered descent abort program (because it simulates lunar ascent)


ThatRedMelon (approved):
All processors tested worked as intended, giving correct results with seemingly accurate trim angles.

PR #1439, opened 2025-05-03
A few RTCC fixes

indy91:
Changes in this PR:
-Add Vector Compare Display error checks and messages on on-line monitor for being below entry interface, maneuver error, vector rotation errrors and other unspecified errors
-Vector Compare Display spherical coordinates (including latitude and longitude) are calculated in Earth/Moon fixed coordinates and not inertial coordinates
-RTCC calculation of AOT counter angles are enforced to be displayed as 0 to 360 degrees
-LOST display declination display with plus sign if the declination is positive
-IU state vectors are saved in platform coordinates and only converted to BRCS coordinates upon being saved in the vector evaluation table. A previous change included the conversion in the second step, but didn't remove it in the telemetry vector saving
-Fix some display centering issues in the RTCC MFD for the two-impulse multiple solution and AGC state vector uplink pages
-Fix minor formatting issue on RTCC MFD Ascent PAD and Abort Scan Table maximum time input
-Initialize lunar impact input for impact time to 0, it could previously show a random time initially

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_launch/rtcc.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_launch/rtcc.cpp
@@ -35331,7 +35331,7 @@ void RTCC::BMSVEC()
 
 	OELEMENTS coe;
 	double mu, r_apo, r_peri, R_E, r, v, lat, lng, gamma, azi, gmt;
-	int numvec = 0;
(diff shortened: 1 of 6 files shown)


ThatRedMelon (approved):
All fixes work as intended

PR #1445, opened 2025-05-18
Coelliptic Sequence Rendezvous Processor overhaul

indy91:
This update brings the processor (sometimes called SPQ) up-to-date with the latest calculation methods in the RTCC.

-Using the state vector struct VehicleDataBlock instead of SV, which allows drag to be taken into account
-New option to define the CDH point which is more compatible with the AGC
-Many new error checks and messages
-Slight MFD page change in the RTCC MFD
-New set of utility functions also to be used by other rendezvous processors in the future
-Delete old utility functions
-Small powered descent abort processor changes to be more clear about GMT vs. GET

Only very few actual new features are in this update. Support for plane change maneuvers and dwell orbits will come in the future.


ThatRedMelon (approved):
Achieved correct SPQ results for both CSI and CDH, both with and without MPT.

PR #1446, opened 2025-05-21
CSM VC: Panel 12 clickspots active in all LEB view positions

indy91:
Previously the panel 12 clickspots were only accessible in the SATVIEW_LOWER_CENTER view position, which is between the center seat and the tunnel. Panel 10 has issues with the MDC switches being clicked in the LEB view positions, but I don't seen any such issues with panel 12. So these are the view positions where the panel 12 clickspots are now active and (still) inactive:

Active:
-G&N panel
-LEB left
-LEB right
-Tunnel
-Lower center (between center seat and tunnel, it was only active in this view positions before this PR)

Inactive:
-Left seat
-Center seat
-Right seat
-Left docking window
-Right docking window
-Side hatch
-Upper center (Above side hatch)


ThatRedMelon (approved):
Could not identify any issues with the view positions the switch is active for.

Issue #1447, opened 2025-05-25
Ascent Rendezvous Monitor Time Formatting Issue


ThatRedMelon:
When the Short ARM is initially loaded, it first tries to format the default TW value as below:

https://github.com/orbiternassp/NASSP/blob/8cb11274ec78bcb3fc8326233be5e4386cebc39c/Orbitersdk/samples/ProjectApollo/src_rtccmfd/ApolloRTCCMFD.cpp#L10115-L10123

Where eventually it tries to format it as HHHMMSSCS using the code below:

https://github.com/orbiternassp/NASSP/blob/8cb11274ec78bcb3fc8326233be5e4386cebc39c/Orbitersdk/samples/ProjectApollo/src_rtccmfd/OrbMech.cpp#L263-L273

However, due to the fact that the Short ARM defaults to DT, the display is meant to show as a negative value, but the time formatting display flips the sign of any negative time, and hence the Short ARM treats the TW as a GET, instead of a DT, and calculations fail. This is only on issue on first load, as this formatting code is not called on subsequent entries.

PR #1448, opened 2025-05-26
RTCC time formatting: Fix negative times being shown as positive if t…

indy91:
…he most significant (e.g. hours in the format hhh:mm:ss) is zero (fix for issue #1447)


ThatRedMelon (approved):
Tested a majority of the RTCC displays with time formatting, both with and without negative inputs showed no unintended results

PR #1450, opened 2025-05-27
Add ORDEAL slew switch keybinds

### LINKED ISSUE #1443: QoL Issue: Add keyboard shortcuts for ORDEAL slew
One issue many have had is in the virtual cockpit, one cannot see the ORDEAL slew switch and the FDAI at the same time unless you zoom way out in FOV. 

A QoL addition would be to add a slew up and slew down keyboard command to the ORDEAL class in order to allow slewing without having to look at both.

ThePuzzlemaker:
Fixes #1443 

Adds `Alt-O` to slew ORDEAL up and `Alt-L` to slew it down.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_csm/saturn.cpp
@@ -3848,6 +3848,22 @@ int Saturn::clbkConsumeBufferedKey(DWORD key, bool down, char *kstate) {
 					bRecovery = true;
 				}
 				return 1;
+			case OAPI_KEY_O:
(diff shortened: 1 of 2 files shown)


Jordan-64:
I have two other suggestions.
1. A custom camera that opens a window where you can see the FDAI when you click the slew button.

2. Define two click points at the top and bottom of the FDAI itself to set the slew value.

rcflyinghokie (changes requested):
Alt-W and Alt-S are the AOT "slow" keybinds, new ones need to be selected

rcflyinghokie (approved):
Works as advertised, #1451 includes updated quick start guide with new keybinds

PR #1452, opened 2025-06-03
Add Skylab Function Control to PAMFD

rcflyinghokie:
This allows the user to see current Skylab lighting and attitude control status as well as toggle lighting on and off and switch attitude control modes without having to switch to Skylab.


ThatRedMelon (dismissed):
All MFD options and MCC changes work as intended

indy91 (approved):

PR #1456, opened 2025-06-12
A few RTCC fixes

indy91:
-A few more checks on ephemeris tables being empty when their left and right limit are written
-Uninitialized memory was used for the docking angle when putting a RTE maneuver on the MPT
-Lunar ascent integrator (when used for ephemeris generation) can use CSM state vector beyond ephemeris limits (useful for prelaunch mission planning)
-Lunar ascent integrator now has an error check on fuel exhaustion (but not processed correctly yet during ephemeris generation)
-Lunar ascent integrator previously returned the negative of the APS fuel used (bad LM mass after insertion)
-Fix LDPP plane change calculations

Suggested test procedure for the two main lunar ascent bugs:

Propellant used for ascent: Put the ascent maneuver on the MPT in any scenario, the MPT display should show a wrong, large DV remaining due to using a negative amount of propellant. Checkout Monitor after ascent shows a wrong LM mass, higher than before liftoff. The PR fixes this.

Wrong CSM state vector: This one is more difficult to test. It has to be a scenario in Earth orbit or before launch. All CSM and LM maneuvers before lunar ascent on the MPT. Then the ascent maneuver should use a bad CSM state vector and show a weird DV remaining, some DMT values might also be wrong. Potentially it runs into a newly added error check and an online monitor message is written. This behavior is also fixed by the PR.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_launch/rtcc.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_launch/rtcc.cpp
@@ -11987,6 +11987,7 @@ void RTCC::LunarAscentProcessor(const LunarAscentProcessorInputs &in, LunarAscen
 	m1 = in.m0;
 	t_total_old = t_total;
 	integ.Init(unit(R));
+	out.ErrorCode = 0;
 
 	while (stop == false)
 	{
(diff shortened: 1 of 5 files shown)


ThatRedMelon (approved):
All fixes work as intended

PR #1458, opened 2025-06-17
Skylab 2 Checklists

GoForPDI:
Based on SL-3 launch checklist, SL-4 rendezvous book, and SLM-1 operational trajectory. Goes through hard docking, no EVA until we get more documentation.


GoForPDI:
Converting to draft because more documentation recently became available

GoForPDI:
Marked as ready for review since the checklists are done - I'd still like me when I get around to it or somebody else who's willing to fly the mission or at least give these a once over before they are merged

indy91 (approved):

PR #1459, opened 2025-06-17
Powered Descent Abort Program upgrade

indy91:


ThatRedMelon (approved):
All calculations (11,12,13+) give correct results in all cases tested, both with and without MPT. Aborts give DHs close to 15, and AGS values are close to irl. MCC also gives correct results.

PR #1463, opened 2025-06-22
Fix include and include guards missing, preventing build.

n7275:


indy91:
There are a few things that aren't implemented in the best possible way in this PR I feel, I am going to put up my own version, also fixing a few build warnings in the MCC!

PR #1464, opened 2025-06-22
Fix Skylab build errors and MCC build warnings

indy91:


ThatRedMelon (approved):
All build warnings and errors experienced in the current release are fixed. Original change from #1462 also works correctly

PR #1468, opened 2025-06-27
Update LM radar self test behavior to better reflect procedures

rcflyinghokie:
- Adds cos 15 multiplier to tape meter slant range in hover position
- Adds mission specific self test values
- Add multiplier to LM-3 to display what was seen in flight on the tape meter (workaround until the actual source is found)

While this PR doesn't address "why" some of these were seen, it adds the correct behavior and would make the procedures properly align rather than using hardcoded Apollo 11 values for every LM. I would like to see it merged with the caveat that if/when better self test/tape meter information comes about we add it.


rcflyinghokie:
Moving to draft for now until we can implement wiring research

rcflyinghokie:
Happy with this as is until we figure the "whys" of all of these differences. As implemented here, this generates expected self test values per mission instead of a single hard coded value

maxq519 (dismissed):

rcflyinghokie:
Thanks to the LM console guide, the proper scaling values were added with correct results. The only outliers that remain are LM-3 and LM-4

indy91 (approved):

PR #1469, opened 2025-06-28
Implement groundtrack digitals and recovery ascending node displays

indy91:


ThatRedMelon (approved):
All new displays work as intended

PR #1472, opened 2025-07-02
Fix deorbit processor REFSMMAT calculation (was off in roll)

indy91:
The Deorbit REFSMMAT as calculated by the deorbit processor was off in roll due to SPS trim gimbal angles. This has now been corrected.


ThatRedMelon (approved):
REFSMMAT calculation completes successfully, giving the correct deorbit angles:

image

rcflyinghokie (approved):

PR #1476, opened 2025-07-07
Improve handling of RTCC cape crossing table

indy91:
The cape crossing table is used with the MPT feature to generate a table of revolution start times. These are the crossings of the Cape Canaveral longitude for Earth, 180 degrees longitude for the Moon.

The improvements are:
-The current revolution is being saved/loaded in such a way that the current rev stays intact during saving and reloading
-The MED to change the current rev (P17, on the MPT display) has been improved and now inputs the rev at present time
-Print error messages from the RTCC function that calculates the cape crossing times


ThatRedMelon (approved):
Improvements work as described

PR #1477, opened 2025-07-10
Implement Apollo 9 third S-IVB burn

indy91:
The most simple implementation I could come up with. The LVDC goes into TB7 at cutoff of the second S-IVB burn like normal. Then at a fixed time since the first TB6 (restart prep for second burn) the LVDC goes into TB6 a second time, utilizing the predicted burn time variables for the second TLI opportunity, which have been loaded with the numbers for the Apollo 9 third S-IVB. The mission scenarios up to the pre SPS-1 scenarios have been updated to allow the third S-IVB burn to happen.


ThatRedMelon (approved):
S-IVB restarts as it should, and all behaviour is as expected.

Care should be taken when implementing the new mission scenarios as these edits will need to be done there too

PR #1479, opened 2025-07-18
Virtual cockpit

Jordan-64:
There are several improvements in the VC, the most important being

Flashing has been added in the VC.
Panel 300-305 has been added in the CM.
Remodeled Circuit Breakers in both VC's.
New 3D model of the SPS Engine Injector Valves.
Added Hotkey (Left CTRL-H) for Arrows which pointing to the Cue Card Clickpoints.
Added simplified Interior of the White Room in LUT Service Arm 9.
Lighting switch in LM Ordeal is now working.
New thumbwheels in CMVC, which are now similar to the originals, have also been removed from the integrated lights, as the numbers on the wheels are black and therefore it makes no sense.
New Side Hatch in CMVC. Also the BPC has now a "Hatch" which is animated with the Side Hatch, so we can have a view in the White Room.
Added Crosspointer lights in LMVC
Added TTCA mesh in LM VC
The round dials in LM VC have been improved.
Improved more accurate round meters in LMVC.
Added Panel 230 in CMVC for all missions not functional for now until the work in 2D is done and merged.
Added working AOT reticle knob with display in LMVC


rcflyinghokie (dismissed):
#include <gcConst.h> does not allow building in a cmake/open orbiter build, the definitions need to be looked at/addressed.

This section should be removed from saturn.h unless there was a reason to include it, in which case it could be moved to saturn.cpp

Added a PR to this branch that removes the code.

rcflyinghokie (commented):
These changes cause lengthy hanging on the loading screen, need to be fixed

rcflyinghokie (commented):
Seems to fix the hang on my custom flight

rcflyinghokie:
Unless anything is broken, and after the interior light additions, I recommend a freeze on VC work so we can go over the code and begin integration

rcflyinghokie (dismissed):
Has been tested in O16 and O24 pretty extensively

indy91 (approved):

rcflyinghokie (approved):

PR #1482, opened 2025-07-28
More reliable convergence on AEG argument of latitude search with high eccentricity orbits

indy91:
Also changes the AEG input element limits to allow all inclinations.

This PR can improve the convergence of General Purpose Maneuver processor height maneuvers in highly elliptical orbits.


maxq519 (approved):
Works great

PR #1483, opened 2025-08-03
Improvements for MED inputs and Saturn IB launch targeting

indy91:
MED input improvements:

-Make MED input pages persistent during an Orbiter session. Previously they reset when leaving and reentering the page.
-MED inputs can now be loaded from a text file on the online monitor page. The file has to be in the main Orbiter folder, and one MED input has to be done per line in the file.
-MED M50 can be input via normal MED input, not just MPT init page
-Add error return from P13 MED
-The LoadLaunchDaySpecificParameters function, which loads launch day init files and initializes the ephemerides, is now called when the P80 MED (launch date) is input. This is a slight departure from how these things actually worked, but we don't have full documentation how this actually fully worked, so I think this acceptable (for now).

Launch targeting improvements:

-Vector time input for Saturn IB launch targeting
-Launchpad selection on input page for Saturn IB launch targeting
-Launch azimuth polynomial coefficients can be loaded in RTCC mission constants file or changed via input

As an example of these improvements, see the attached scenario, RTCC constants file and a MEDInput file. The scenario is a Saturn IB dual launch mission and the MED input file initializes the MPT (P80, P10, M55, M50, M51) and inputs the first launch as an insertion state vector to the MPT (P13), so that launch planning for the second Saturn IB launch can be started immediately.

[Skylab 2 As Gemini.scn.txt](https://github.com/user-attachments/files/21566988/Skylab.2.As.Gemini.scn.txt)
[Skylab 2 As Gemini.txt](https://github.com/user-attachments/files/21566989/Skylab.2.As.Gemini.txt)
[MEDInput.txt](https://github.com/user-attachments/files/21566990/MEDInput.txt)


ThatRedMelon (approved):
All features work as intended

PR #1484, opened 2025-08-05
Improve connecting/disconnecting of Saturn/launchpad umbilical classes in various cases like pad aborts or vessels being deleted

indy91:
This PR changes the connecting/disconnecting behavior of the umbilical classes for the IU, S-IB and S-IC to work more like our general connector class.

The reason for this change is an (apparent) different deletion order of vessels in Open Orbiter which causes a crash of the simulation on exiting it. This PR also deletes some special code handling pad aborts where the disconnection of the umbilical isn't caused as usual by a call of the Disconnect function. With this change the order of deletion of the umbilical/systems classes and the source of the disconnect should not matter at all.

Testing for this PR should all be possible with scenarios shortly before launch. If the Saturn launches normally then the umbilical connect (on scenario load) works correctly. Disconnects can be tested with normal launches and pad aborts. There should be no crash on scenario exit. There wasn't any crash in the Orbiter Beta anyway. This PR is mostly for Open Orbiter and a good cleanup of some bad workarounds.

This change is also mostly temporary as the WIP branch for Saturn V docked stages (#1382) actually changes the umbilicals to use the connector class.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_launch/IUUmbilical.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_launch/IUUmbilical.cpp
@@ -31,34 +31,29 @@ IUUmbilical::IUUmbilical(IUUmbilicalInterface *ml)
 {
 	IuUmb = ml;
 	iu = NULL;
-	IUUmbilicalConnected = false;
 }
 
 IUUmbilical::~IUUmbilical()
 {
+	Disconnect();
 }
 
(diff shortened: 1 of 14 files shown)


ThatRedMelon (approved):
The sim doesn't crash in O16 and O24, as expected

rcflyinghokie (approved):
No crashes here in OO

Issue #1487, opened 2025-08-08
AGS Not controlling properly with AGS AC cb pulled


rcflyinghokie:
Looking into the AGS, it seems the AGS AC breaker allows the AEA to generate the attitude and error signals that go to the ordeal and FDAI. However it doesn't appear it should prevent the AGS from properly steering as that power should be what seems to be from the ATCA. 

Procedurally, this was found as Apollo 10 has the AGS AC cb pulled before APS depletion, and there was an AGS switchover. In NASSP, this makes the LM slowly drift from attitude while burning on AGS as if its chasing "frozen" error signals, and I believe this is the incorrect behavior.

rcflyinghokie:
#1488 fixes this

PR #1488, opened 2025-08-10
AGS attitude error output is generated with power from the STAB/CONT - ATCA circuit breaker, not the AC BUS B - AGS circuit breaker (fixes issue #1487)

indy91:


rcflyinghokie (changes requested):
needs to look for minimum DC voltage not AC voltage on the ATCA breaker

rcflyinghokie (approved):
Works correctly!

PR #1490, opened 2025-08-13
RTCC: When DKI and SPQ maneuvers are moved to the MPT, all maneuvers are moved instead of just the first one

indy91:
Also adds inputs for Two Impulse, DKI and SPQ transfer pages for all maneuvers. Lunar Descent Planning processor does not move all maneuvers to the MPT at once yet as I don't understand how that would work in some cases.

EDIT: Not quite sure I am happy with the new maneuver transfer page when used in non-MPT mode. Aside from that, this PR is functionally ready.


ThatRedMelon (approved):
New functions work as described

PR #1496, opened 2025-08-27
RTCC: Fix transfer to MPT of LOI/MCC maneuvers; save/load entire LOI initialization table in scenarios

indy91:


ThatRedMelon (approved):
Saving and loading works correctly, bugfix works as intended.

PR #1501, opened 2025-09-04
Further fixes for umbilical disconnect (causing CTD when closing an Orbiter sim session)

indy91:
Testing involves any Saturn V scenario before launch. Random crashes can happen in TSMUmbilical::Disconnect because the SICSystems "sic" has already been deleted but the pointer has not been nulled, so the TSMUmbilical::Disconnect code will try to do "sic->TSMUmb = NULL;", which can cause an access violation.


ThatRedMelon (approved):
Tested a bunch of scenarios on the launchpad, everytime sim was quit it went back to the orbiter launchpad without crashing

PR #1502, opened 2025-09-04
RTCC landing site coordinates are initialized from skeleton flight plan interpolation (MED F62) instead of from the RTCC init files

indy91:


ThatRedMelon (approved):
Features work as intended

PR #1503, opened 2025-09-11
Make the RTCC reentry simulation more realistic by adding roll dynamics and attitude control calculations

indy91:
The impetus for this update was a prediction by the Apollo 8 Entry PAD that the reentry guidance would go into P65. Apollo 8 flew a trajectory very close to getting into P65, but it didn't on the actual mission. The RTCC reentry simulation before this updated allowed the CM to instantly achieve a desired roll rate, not limited by the amount of angular acceleration that the CM RCS could provide. The real RTCC reentry simulation went as far as simulating the Entry DAP, which makes it a quite realistic simulation. This update is about halfway there. Not a full DAP simulation, but enough attitude control logic and roll dynamics being simulated to avoid the unrealistic P65 prediction.

As I have gone through the whole simulated entry guidance with this update there are also lots of small improvements like initializating all flags properly and protections against negative values in square roots etc. These protections were mentioned in the RTCC Requirements document.

The best testing of this update is probably any calculation that uses the reentry simulation, like Entry PAD calculations, RTE Digitals and the deorbit targeting.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_rtccmfd/ReentryNumericalIntegrator.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_rtccmfd/ReentryNumericalIntegrator.cpp
@@ -62,7 +62,7 @@ GNDataTab::GNDataTab()
 
 ReentryNumericalIntegrator::ReentryNumericalIntegrator(RTCC *r) : RTCCModule(r)
 {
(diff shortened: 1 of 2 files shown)


ThatRedMelon (approved):
From my tests, predictions seem to be accurate, with no issues.

PR #1504, opened 2025-09-13
Add vector panel summary entry capability; fix a few AGOP inputs

indy91:
Add capability to enter state vectors on the vector panel summary display:

image

Also fixes a few AGOP inputs, mostly for option 4 (antenna pointing).


ThatRedMelon (approved):
Manual entry works as intended in multiple vector locations. AGOP angles provide correct angles to acquire ATS manually

PR #1506, opened 2025-09-16
RTCC: Better support for CSM + LM ascent stage center of gravity calc…

indy91:
…ulations; generate error for non-elliptical orbit in general purpose maneuver processor


ThatRedMelon (approved):
Calculations appear to work as intended

PR #1507, opened 2025-09-18
Change LM engine start button to be momentary

indy91:


rcflyinghokie (approved):
Visual and spacecraft behavior works as expected, also with checklist MFD.

PR #1510, opened 2025-09-27
MPT Docking Maneuvers

indy91:


ThatRedMelon (approved):
Docking maneuver successfully adds to the MPT without noticeable issue

PR #1512, opened 2025-09-29
Change attitude commands to go through RHC2

ThatRedMelon:
As per the 1969 AOH below, the left couch used RHC 2, instead of 1. This PR changes command inputs to be through RHC2 instead of 1, for procedural simplicity.

image


rcflyinghokie (approved):
Looks good here, simple swap of the "LH seat" RHC

PR #1518, opened 2025-10-10
Fix Apollo 9 AEA Pad Count

ThatRedMelon:
Previously it was 1 too few and was losing the 703 address


rcflyinghokie (approved):
This was on me when I added the ullage! Approved!

PR #1519, opened 2025-10-10
Show decimal and octal star codes of additional RTCC stars in marker file

indy91:
New format:

image


rcflyinghokie (approved):
They look good!

PR #1521, opened 2025-10-11
Convert LGC and AEA padloads to vectors and remove requirement to load LMPADCNT and AEAPADCNT in scenarios

indy91:
Testing:

-LGC and AEA padload properly save/load in Saturn class
-At CSM/LV separation, padload is given through the S-IVB class to the LM class
-Padloads end up as EMEMs in the newly created LM class


ThatRedMelon (approved):
Scenarios load and save correctly

rcflyinghokie (approved):
Looks good here

Issue #1401, opened 2025-04-04
LM Track Mode Switch Behaviour May Be Incorrect


ThatRedMelon:
As discussed on discord, the behaviour when switching the track mode switch to OFF may be incorrect. Some debate has ocurred on whether or not signal should be lost in this position or not, and more documentation is necessary in figuring out the correct behaviour. The current consensus seems to be that RF signal can still be received, regardless of if the steerable antenna package has power or not, but this is not confirmed.

rcflyinghokie:
Looking at this, I think you are correct, the track mode switch appears only to energize the servos, it doesn't stop signal from flowing.

rcflyinghokie:
I think just removing this from the LEM_SteerableAnt::Timestep would be safe to do:

```
	if (!IsPowered())
	{
		SignalStrength = 0.0;
		return;
	}
```

Still need to investigate more, but it seems if the electronics package is turned off, it doesnt/shouldnt inhibit the signal strength

rcflyinghokie:
<img width="1925" height="979" alt="Image" src="https://github.com/user-attachments/assets/df586e8b-9e50-43ad-9a4e-87f37535007c" />

Would the RF signal flow properly if those lines with X's had no power

PR #1529, opened 2025-10-16
LM S-BD Steerable Updates

rcflyinghokie:
- Remove lack of signal strength if unpowered (RF should run through independent of the electronics package, also moving switch from AUTO to SLEW through OFF would cause a drop in signal)

- Remove "cardioid" from antenna so it evenly heats and cools (electronics are omnidirectional)

- Base movement power draw/heat generation on drive rates


rcflyinghokie:
#1401

indy91 (approved):

PR #1533, opened 2025-10-25
Add Apollo 5 checklist

indy91:
In detail:

-Move T-1 hour scenario to supported missions
-Adds Apollo 5 LM checklist file with prelaunch procedures as well as a guide through the all automatic nominal mission
-Updated T-60 seconds scenario
-Remove special unmanned LM code that would force the checklist to auto execute
-LM Programer can turn on primary S-Band transceiver and switch from descent to ascent H2O supply


maxq519 (approved):
Tested in OpenOrbiter, works as intended

PR #1534, opened 2025-10-26
Bugfix in RTCC longitude search routines and remove Entry PAD check splashdown latitude is exactly zero

indy91:


ThatRedMelon (approved):
All seems to work as intended

PR #1535, opened 2025-10-27
Apollo 12 ALSEP

TINYBOB1:
After a long delay (oops...) the ALSEP package for Apollo 12 is ready.  Deploys in the order from the Lunar Surface Checklist, and for the appropriate crew member.  No code for pointing antennae at earth is in yet, but that would be a nice feature.  The meshes are setup for easy rotation for the appropriate mesh groups.  Thanks to @maxq519 for the EASEP code which I "borrowed"


TINYBOB1:
12 ALSEP
12 SBnd
ALSEP deployed (sadly under parts of the surface since the "physical" surface and the displayed ones dont line up)  and the erectable S-Band Antenna

TINYBOB1:
how do we want to tell users how this works?  we dont have a checklist system for EVA

rcflyinghokie:
> how do we want to tell users how this works? we dont have a checklist system for EVA

I have lists in the QSG #1541

rcflyinghokie (dismissed):
Seems to work correctly here

indy91 (changes requested):
There are separate "TvPlanted" and "ColorTvPlanted" variables in the LEVA class now. ColorTvPlanted seems to be unused. Maybe it simply can be deleted? That is all I found, when that is resolved this PR can be merged.

indy91 (approved):

PR #1538, opened 2025-11-05
RTCC: Use Vector Panel Summary vectors for maneuver calculations

indy91:
This PR adds the capability to most maneuver processors to optionally use a vector from the Vector Panel Summary display as input to the maneuver calculation. Both MPT and non-MPT mode are supported.

Changes in this PR:

-Two Impulse Multiple Solution: Chaser and target vector IDs as optional inputs.
-Two Impulse Corrective Combination: Chaser and target vector IDs as optional inputs.
-Coelliptic Sequence processor (SPQ): Change input format because the MFD page ran out of buttons. Chaser and target vector IDs as optional inputs.
-DKI processor:  Chaser and target vector IDs as optional inputs (only works with MPT for now).
-GPM processor: Add Vector ID as optional input. VehicleDataBlock as input instead of state vector and mass, area separately. But this is just a function interface change, nothing functional. A couple of changes in the RTCC file for the MCC to accomodate this change
-TLMCC processor: Add Vector ID as optional input
-LOI processor:  Add Vector ID as optional input
-RTE processor: Change input method for RTE constraints and add vector ID selection (useable vector table only).
-Lunar Descent Planning processor: Change input format and add vector ID selection
-LLTP not supported yet, will get separate input page
-Perigee Adjust processor: Add Vector ID as optional input


ThatRedMelon (approved):
Vector IDs tested across multiple displays, giving acceptable looking results based on selected ID

PR #1543, opened 2025-11-09
Fix two MPT bugs with executed maneuvers; also fix a RTCC MFD bug

indy91:
- When an executed maneuver is being history deleted its trajectory update number was copied (as part of a struct of common data) from the maneuver data to the MPT header. Executed maneuvers are not integrated through with new trajectory updates, so the stored trajectory update number could be a lower number than the current trajectory update. This could lead to inconsistent numbers between tables (MPT vs. ephemeris table) which leads to errors on certain displays (e.g. error 3 on the checkout monitor). This has been corrected.
- The lunar stay begin and end times were not reset on trajectory updates. If there was an executed powered descent maneuver on the MPT (on the surface at present time), that maneuver is being deleted and a new on-orbit trajectory update is performed, the lunar stay begin/end times were still as if the powered descent is on the MPT. Now these times are reset/regenerated on every trajectory update.
- Fix Two Impulse transfer page input marker
- Show N/A instead of blank in the PrintUllage function

Terminology:

Executed maneuver = maneuver on the MPT that has already been executed, i.e. its time of ignition is before present time
History deletion = deletion of executed maneuver

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_launch/rtcc.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_launch/rtcc.cpp
@@ -16703,7 +16703,7 @@ void RTCC::PMMFUD(int veh, unsigned man, int action, std::string StationID)
 			PMXSPT("PMMFUD", 42);
 			return;
 		}
-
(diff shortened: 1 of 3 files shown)


ThatRedMelon (approved):
No issues experienced with this update

PR #1547, opened 2025-11-15
Make TPI times page compatible with the MPT and some RTCC bugfixes

indy91:
- TPI times page is now compatible with the MPT and uses the input guess time as the vector time
- Vector Panel Summary display properly refreshes even if present GMT is negative
- PDI PAD has better MPT compatibility and uses landing minus 20 minutes as vector time
- Fix SPQ vector time input display overlapping
- VECPOINT angles displayed in all calculation modes
- Show 3 instead of 2 decimal places of REVS1 on TLMCC and LOI inputs
- Add GPM vector time input (and reorganize the order of the display for it)


ThatRedMelon (approved):
TPI times page seems to work with MPT vectors

PR #1551, opened 2025-11-23
Implement interface to the erectable antenna in LM S-Band code

indy91:
Should have been merged as part of the ALSEP update, but was forgotten about.


rcflyinghokie (approved):
signal strength correctly disconnected

PR #1553, opened 2025-11-24
Fix signal conditioner source of LM S-Band receiver signal strength on telemetry

indy91:


rcflyinghokie (approved):
Correct values are now displayed

PR #1557, opened 2025-12-06
Add code for handling P30 REFSMMATs with MPT enabled; delete old current REFSMMAT indicator

indy91:
The logic with MPT enabled first generates a DMT REFSMMAT (using MED logic, either G11 for the CSM, or G21 for the LEM) and then moves the DMT REFSMMAT to the CUR REFSMMAT with MED G00.

Instead of TIG and DV as input and the old REFSMMATCalc function being used this new MPT logic now just takes a maneuver number as input and acts as if these two MEDs were called.


rcflyinghokie (approved):
Works correctly in my LOI-1 case

PR #1558, opened 2025-12-07
Add AOT Zero Detent

rcflyinghokie:


rcflyinghokie:
This is fully functional but doesn't have any 2d visualization on the detent position yet. @Jordan-64 is working on the VC side.

This can be merged as is with the debug lines, or we can rely on just the VC visual indication, or we can add a 2d indicator on the AOT bmp as well. I am open to any of these solutions.

rcflyinghokie:
bmp added and functional, last hurdle is making it so the clickspot doesn't always toggle the switch position. The keybind seems to use the function code that prevents it from being pushed in when the reticle is > 1 degree from center but the click ignores this if somehow

Jordan-64 (approved):
In my tests, it worked as expected.

indy91 (approved):

PR #1567, opened 2026-01-07
Fix burn parameters type 3 (IMU) and 4 (LVLH) for MPT direct input maneuvers

indy91:


rcflyinghokie (approved):
IMU and LVLH external inputs to MPT seem to generate the expected attitudes in the maneuver pad

PR #1582, opened 2026-01-28
RTCC: Implement star sighting table display

indy91:


rcflyinghokie (approved):
Tested a few ground and celestial pointing targets with good results.

PR #1598, opened 2026-02-20
RTCC MFD fixes

indy91:
-Combine LDPP with SPQ/DKI maneuver transfer in non-MPT mode 
-Simplify vector panel summary inputs

Things to test:
-The LDPP transfer page used outdated code (and didn't output ullage data for the Maneuver PAD). It has now been combined with the SPQ/DKI transfer logic, like it works in MPT mode. So testing should involve LDPP, SPQ and DKI transfers.
-The DC buttons on the Vector Panel Summary now default to the selected CSM/LM names, instead of automatically using these selected vessels if the input is left blank. TUP button now defaults to "DC". Testing should involve checking if these buttons still do what they are supposed to.

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_launch/rtcc.h
+++ b/Orbitersdk/samples/ProjectApollo/src_launch/rtcc.h
@@ -3990,8 +3990,8 @@ public:
 	{
 		EphemerisData sv_man_bef[4];
 		VECTOR3 V_man_after[4];
-		int num_man;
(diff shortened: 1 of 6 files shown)


rcflyinghokie (approved):
Buttons work as expected, prepopulating the vessel and TUP with DC. Tried the lunar descent and SPQ as well seemed to transfer to maneuver pads. Did not get to test a DKI.

PR #1599, opened 2026-02-22
RTCC fixes and improvements

indy91:
-Use more up-to-date functions for EntryUpdateCalc
-Save weights at maneuver begin and main engine on in MPT
-Fix MPT bug where deleting an executed undocking maneuver does not update the total initial weight to the undocked value

Testing: Splashdown update in RTCC MFD (for the EntryUpdateCalc function) should work like before and this specific MPT condition should be tested, if it now works correctly.


rcflyinghokie (approved):
tested the splash update functions, works as expected.

Issue #740, opened 2022-03-11
Battery C Charge Indicator Goes To Zero Upon 2D Panel Switch


rcflyinghokie:
If Battery C needs to be charged, and the charger is turned on, it indicates charging amperage. However if you switch panels in 2D and return to MDC3 it reads zero. You can turn off the charger and back on and the current comes back up but the behavior is reproducible.

The behavior does not exist in the VC or when going from external to 2D view. Additionally it appears looking at FC flows that the charging stops after the panel switch, not just the display itself.

n7275:
Well this is a weird one. Good find. Somehow I missed the IRC discussion on this if there was one, but I'm assuming that that there is a refresh in the wrong place?

indy91:
There is some wiring that is right now done every time you switch between 2D panels. And with battery C there is some re-wiring done in code when the battery charger is charging it. That is because our systems simulation doesn't allow two-way power flow and there are some CBs that actually have two-way power flow if battery C is used vs. battery C being charged. I'll work on this issue very soon, it's not a difficult fix.

n7275:
> There is some wiring that is right now done every time you switch between 2D panels. And with battery C there is some re-wiring done in code when the battery charger is charging it. That is because our systems simulation doesn't allow two-way power flow and there are some CBs that actually have two-way power flow if battery C is used vs. battery C being charged. I'll work on this issue very soon, it's not a difficult fix.

I have some Matlab code I wrote a while back for drawing from multiple power sources, if it's helpful. https://gist.github.com/n7275/8c95469fbe2d7f234eded87c79130c4a

indy91:
Fixed by #1605

PR #1605, opened 2026-03-04
Move initial wiring for battery C charging from SetSwitches to System…

indy91:
…sInit, so that it does not get reset upon reloading of the 2D panel (fix for issue #740)


ThatRedMelon (approved):
Fix works as intended

PR #1608, opened 2026-03-07
LM CO2 partial pressure sensor has realistic voltage curve; consisten…

indy91:
…t display on meter in both 2D panel and VC; indicated partial pressure a function of total pressure; component light uses relay in CWEA


rcflyinghokie:
Works well for me, TM client needs something to go along with it though since it should get the same signal as the internal LM components

indy91:
https://github.com/indy91/LMTelemetryClient2/commit/78be5324884a11c2a1d833f3b2b98935aef5d8dd

rcflyinghokie (approved):

PR #1611, opened 2026-03-12
RTCC fixes

indy91:
-REFSMMAT uplink page shows load number and GET of generation
-Fix error message formatting on REFSMMAT uplink page
-Update of anchor vector to present time before a trajectory update is allowed to go through entry interface and only fails at the kickout altitude


rcflyinghokie:
Checked the REFSMMAT page improvements, nice to have!

ThatRedMelon (approved):
Features work as intended, with some nice QoL!

PR #1612, opened 2026-03-13
LM VC: Fix clicking through ECS panel from LMP position; allow clicking on forward hatch except in the DSKY view position

indy91:
Changes in this PR:

-Forward hatch relief valve clickspot can now be used in all LM view positions except in the DSKY view (LMVIEW_DSKY). Previously it could only be used in the forward hatch view position (LMVIEW_FWDHATCH).

-The following switches can now not be used anymore in any of the three LMP view positions. The three positions are LMVIEW_LMP (by default looking at panel 2), LMVIEW_CBRIGHT (looking at the panel 16 circuit breakers) and LMVIEW_ECS2 (looking at the oxygen control module):

-Suit Circuit Relief valve
-Cabin Gas Return valve
-Ascent H2O valve
-Water Separator Selection valve
-CO2 Canister Selection valve
-Primary CO2 Canister Open/Close selector
-Primary CO2 Canister Vent pushbutton
-Secondary CO2 Canister Open/Close selector
-Secondary CO2 Canister Vent pusbutton

### DIFF
--- a/Orbitersdk/samples/ProjectApollo/src_lm/lemvc.cpp
+++ b/Orbitersdk/samples/ProjectApollo/src_lm/lemvc.cpp
@@ -1358,17 +1358,34 @@ void LEM::RegisterActiveAreas()
 
 	// ECS Panels
 
-	oapiVCRegisterArea(AID_VC_ROT_SUITCIRCUITRELIEF, PANEL_REDRAW_ALWAYS, PANEL_MOUSE_DOWN);
-	oapiVCSetAreaClickmode_Spherical(AID_VC_ROT_SUITCIRCUITRELIEF, Rot_SuitCircuitReliefLocation + ofs, 0.02);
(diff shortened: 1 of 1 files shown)


rcflyinghokie (approved):
I will have to say I dislike not being able to use the DSKY in the LMP view, especially when doing DSKY/DEDA work. Is there any way to make only the hatch valve usable and not the hatch handle? If not I would say revert this change as its more practical to access DSKY and DEDA in this view.

rcflyinghokie (changes requested):
I will have to say I dislike not being able to use the DSKY in the LMP view, especially when doing DSKY/DEDA work. Is there any way to make only the hatch valve usable and not the hatch handle? If not I would say revert this change as its more practical to access DSKY and DEDA in this view. I can press some of the DSKY buttons but keep toggling the hatch handle.

rcflyinghokie (approved):
Current version allowing just the vent valve from the LMP view works well, no more interruption of the DSKY in that view (which is important for DSKY/DEDA work)

PR #1613, opened 2026-03-13
FIX CMVC FDAI ball Pitch deviation

Jordan-64:


rcflyinghokie (approved):
Seems to fix the FDAI position error that existed before

PR #1618, opened 2026-03-23
LM Interior Lighting Improvements

rcflyinghokie:
- Change internal meter text backlighting to never be zero, prevents "no power" off from being dimmer than "not selected" off
- Move voltage division to LCA class
- Add floodlight outputs


Jordan-64:
I haven't tested it, but the changes make sense. Previously, it didn't take into account the possibility of no power, and the text was completely darkened.

PR #1621, opened 2026-04-13
The second possible solution (with worse Delta V) for an apsides change in the General Purpose Maneuver processor can be selected; fix MCC loading bug

indy91:


maxq519 (approved):
General Purpose Maneuver apsides change solutions work as intended

PR #1626, opened 2026-04-26
Add Talkback sounds

Niker25:
This update aims to implement a sound when the talkback status is gray or barbepole, depending on the current state of the animation (0 or 3).

Sounds source:
https://www.youtube.com/watch?v=Qq7iOpYUqjM


rcflyinghokie:
Does this apply to the LM talkbacks as well in this code? They would not sound the same I don't think since they have a flag almost on a rotating axis instead of a shutter.

image

image


It might be worth making a class for those different ones so the rotational animation can be created as well in the future.

rcflyinghokie:
Since there has been no commentary or changes, I am going to make this a draft

PR #1635, opened 2026-05-09
Add permanent saving of RTCC MFD module data in scenario

indy91:
Scenarios permanently save parameters now whenever the RTCC MFD module is activated in the Orbiter launchpad. These parameters currently are:

-MPT active/inactive flag
-CSM and LM vessel names (previous saved in RTCC class)
-Time of lunar liftoff
-Sextant star check time for Maneuver PADs
-TPI time

Every other kind of RTCC MFD saving that previously existed (but only worked when the MFD was open at scenario exit on the 2D panel or glass cockpit) has been removed.


rcflyinghokie (approved):
Seems to work as advertised

PR #1636, opened 2026-05-20
Implement RTCC expendable weights table

indy91:
The RTCC expendable weights table is a table used to define discrete weight losses and gains for the MPT that are independent of any propulsive maneuvers. One good example would be the SIM bay door jettison on Apollo 15 to 17.


maxq519 (approved):
Works as intended

PR #1642, opened 2026-06-04
Most RTCC MFD input boxes now have a default input string

indy91:


rcflyinghokie (approved):
Took a little to get used to, but I think its quite useful

PR #1663, opened 2026-07-05
Change order of LM lights timesteps so that component lights have the SCEA voltages available on the first timestep

indy91:


rcflyinghokie (approved):
Prevents the component light on first timestep as intended

PR #1665, opened 2026-07-09
Checklist MFD: Checklist Navigation page does not crash anymore when no checklist file is loaded

indy91:


rcflyinghokie (approved):
Prevents crash in non checklist condition when NAV is selected

PR #1675, opened 2026-08-27
Remove kill rotation from PAMFD GNC, reorder buttons in checklist MFD

rcflyinghokie:
- Remove kill rotation from PAMFD GNC
- Reorder AUTO EXEC FLASH to better align with use and display in checklist MFD


ThatRedMelon (approved):
The functionality of this PR works as intended. Good to merge if we are in agreement about removing the feature

rcflyinghokie:
I left the function in the code so it can be added back if someone needs it for debugging, beyond that I see no reason to keep a kill rotation button on an MFD in a main distribution.

maxq519 (approved):

PR #1680, opened 2026-09-04
RTCC LM Descent Planning Processor: Better initial guess for TLAND in DOI maneuver calculation if PDI is not at perilune (like Apollo 17)

indy91:
Current behavior: The DOI calculation in the LDPP calculates a preliminary time of landing that assumed PDI happens at perilune. Depending on the calculation option this TLAND is then used to calculate a more precise PDI and TLAND. When this has been converged the two times are biased for the difference between perilune and PDI position using the two angles for this on the LDPP init page. If PDI does not happen at perilune like on Apollo 17 the initial TLAND can be off by enough time that the more precise time cannot converge. This results in an error 32, "Could not converge on landing site".

Changed behavior: The DOI calculation itself already outputs the PDI and landing times using the two angles from the init page, making this initial guess much more precise and the PDI time calculation should converge better.


rcflyinghokie (approved):
ran with some of the scenarios we had issues with and worked out solutions, I see no harm in this implementation

